PR #381, opened 2021-03-24
add chinese currency

mmilenkovic:


src/utils/currencies.js
```diff
@@ -12,9 +12,9 @@ const CurrencySymbol = {
     symbol: '€',
     code: 'EUR'
   },
-  CNY: {
+  RMB: {
     symbol: '¥',
-    code: 'CNY'
+    code: 'RMB'
```
AlbertoElias: Is this the code supported by the api we use?

elias-garcia: `CNY` is the proper one

elias-garcia (approved):

AlbertoElias (approved):

src/utils/currencies.js
mmilenkovic: So international and api supported is `CNY` I'm waiting for the confirmation on what the display value should be. But anyway I've reverted this to `CNY`.

PR #382, opened 2021-03-24
support-ethereum-addresses-in-the-qr-scanner

mmilenkovic:


src/views/shared/qr-scanner/qr-scanner.view.jsx
```diff
@@ -32,7 +32,7 @@ function QRScanner ({ hideMyCode, onSuccess, onError, onClose }) {
    * @returns {void}
    */
   function handleQRScan (result) {
-    if (result && Addresses.isHermezEthereumAddress(result)) {
+    if ((result && Addresses.isHermezEthereumAddress(result)) || (result && Addresses.isEthereumAddress(result))) {
```
elias-garcia, line 35: I would write the if as: `result && (isHermezEthereumAddress(result) || isEthereumAddress(result))`

elias-garcia:
I think that this is not enough to support Ethereum addresses. When you are redirected to the form, the Ethereum address should be converted to a Hermez address (with the prefix at the beginning) and I think that with this change only it will set the input to just the Ethereum address. It won't display the **** in the middle of the address neither.

In the `transaction-form.view` component, we need to remove the L60 and L61 and call `handleReceiverInputChange` function instead, which is already in the same component. This should make it work as expected I think.

src/views/shared/qr-scanner/qr-scanner.view.jsx
mmilenkovic: Done.

mmilenkovic:
> I think that this is not enough to support Ethereum addresses. When you are redirected to the form, the Ethereum address should be converted to a Hermez address (with the prefix at the beginning) and I think that with this change only it will set the input to just the Ethereum address. It won't display the **** in the middle of the address neither.
> 
> In the `transaction-form.view` component, we need to remove the L60 and L61 and call `handleReceiverInputChange` function instead, which is already in the same component. This should make it work as expected I think.

Done.

elias-garcia (approved):

PR #383, opened 2021-03-24
Substract fee from pending txs

elias-garcia:


AlbertoElias (approved):
Can you change the pr to make it to develop?

PR #403, opened 2021-03-29
go-to-account-details-after-sending-transaction-from-home

mmilenkovic:


src/views/transaction/components/transaction-form/transaction-form.view.jsx
```diff
@@ -367,6 +368,7 @@ function TransactionForm ({
         const transactionFee = getFee(fees, true)
         return onSubmit({
           amount: amount,
+          from: { accountIndex: account.accountIndex },
```
AlbertoElias, line 371: Isn't this only necessary for transfers and other txs should go to the home page?

src/views/transaction/transaction.view.jsx
```diff
@@ -60,7 +60,7 @@ function Transaction ({
   const classes = useTransactionStyles()
   const { search } = useLocation()
   const urlSearchParams = new URLSearchParams(search)
-  const accountIndex = urlSearchParams.get('accountIndex')
+  let accountIndex = urlSearchParams.get('accountIndex')
```
AlbertoElias, line 63: maybe use useState?

elias-garcia: Yep, I would use `useState` instead of variable reassignment here

src/views/transaction/transaction.view.jsx
```diff
@@ -339,7 +343,7 @@ const mapDispatchToProps = (dispatch) => ({
   onGoToTransactionOverviewStep: (transaction) =>
     dispatch(transactionActions.goToReviewTransactionStep(transaction)),
   onFinishTransaction: (transactionType, accountIndex, redirectTo) => {
-    if (transactionType === TxType.Withdraw) {
+    if (transactionType === TxType.Withdraw || transactionType === TxType.Deposit) {
       if (redirectTo === WithdrawRedirectionRoute.Home) {
```
elias-garcia, line 347: I think that we don't need this check anymore right? Because the `redirectTo` variable just contains information from where the user comes from (from the `home` view or from the `account-details` view).

This is useful when the user clicks the close button to the redirect him to the view he comes from. But I think that now we don't need to take this into account after the transaction has been finished and just redirect the user to the `account-details` view if he is making a L2 transaction, and to the `home` view otherwise.

Fix me if I'm wrong Miros! Just trying to wrap my head around what we currently are doing and this new flow 😄

src/views/transaction/components/transaction-form/transaction-form.view.jsx
elias-garcia: Yep, I also think that we don't need the `accountIndex here`

AlbertoElias (changes requested):
Lgtm if you remove all the lingering comments

src/views/transaction/transaction.view.jsx
mmilenkovic: Fixed in a different way.

AlbertoElias (approved):

PR #406, opened 2021-03-30
Fixes form to avoid negative balances

AlbertoElias:


src/views/transaction/components/transaction-form/transaction-form.view.jsx
```diff
@@ -162,6 +162,24 @@ function TransactionForm ({
     return showInFiat ? Number(amountFiat.toFixed(2)) : Number(getTokenAmountString(amount, account.token.decimals))
   }
 
+  /**
+   * A precise calculation of the leftover amoutn after substracting fee
```
elias-garcia, line 166: Just a typo in `amount`

elias-garcia (approved):

PR #411, opened 2021-04-06
Clean up pending withdraws in storage automatically in home and accou…

AlbertoElias:
…nt details

### DIFF
--- a/src/store/global/global.thunks.js
+++ b/src/store/global/global.thunks.js
@@ -6,7 +6,7 @@ import { TxType, TxState } from '@hermeznetwork/hermezjs/src/enums'
 
 import * as globalActions from './global.actions'
 import { LOAD_ETHEREUM_NETWORK_ERROR } from './global.reducer'
-import * as fiatExchangeRatesApi from '../../apis/fiat-exchange-rates'
+// import * as fiatExchangeRatesApi from '../../apis/fiat-exchange-rates'
 import * as hermezWebApi from '../../apis/hermez-web'
 import * as storage from '../../utils/storage'
 import * as constants from '../../constants'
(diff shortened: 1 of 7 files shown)


src/store/global/global.thunks.js
```diff
@@ -83,7 +83,15 @@ function fetchFiatExchangeRates (symbols) {
   return (dispatch) => {
     dispatch(globalActions.loadFiatExchangeRates())
 
-    return fiatExchangeRatesApi.getFiatExchangeRates(symbols)
+    return Promise.resolve({
```
mmilenkovic, line 86: Wanna add a `TODO` here, so we don't forget to fix it?

src/store/global/global.thunks.js
```diff
@@ -240,6 +247,37 @@ function checkPendingDelayedWithdraw (exitId) {
...
+      return CoordinatorAPI.getExit(pendingWithdraw.batchNum, pendingWithdraw.accountIndex)
+        .then((exitTx) => {
+          if (exitTx.instantWithdraw || exitTx.delayedWithdraw) {
+            dispatch(removePendingWithdraw(pendingWithdraw.id))
+          }
+        })
+        .catch(() => {})
+    }))
+      .then(() => dispatch(globalActions.checkPendingWithdrawalsSuccess()))
+      .catch(() => dispatch(globalActions.checkPendingWithdrawalsSuccess()))
```
mmilenkovic, line 277: Why success is returned in both cases?

src/store/global/global.thunks.js
AlbertoElias: I'll create an issue

src/store/global/global.thunks.js
AlbertoElias: It's fine if it fails, this is just an automated check to clean up pending withdrawals. I copied the structure from the funcion `checkPendingDeposits` that Elias did

mmilenkovic (approved):

PR #414, opened 2021-04-07
prevent double click on transactions Send button

mmilenkovic:


src/views/transaction/components/transaction-overview/transaction-overview.view.jsx
```diff
@@ -98,6 +99,8 @@ function TransactionOverview ({
         return onExit(amount, account, fee)
       }
       default: {
+        setIsButtonDisabled(true)
```
AlbertoElias, line 102: You would need to do this at the start of the function, or it will only be applied for transfers

mmilenkovic: Makes sense.

AlbertoElias (changes requested):
Please remove the package-lock changes from this pr

AlbertoElias (approved):

PR #438, opened 2021-04-21
Bug/copying bjj copies hez address

mmilenkovic:


src/views/shared/transaction-info/transaction-info.view.jsx
```diff
@@ -68,6 +70,7 @@ function TransactionInfo ({
     case TxType.TransferToBJJ:
     case TxType.TransferToEthAddr: {
       if (accountIndex === txData.fromAccountIndex) {
+        console.log('txData: ', txData)
```
elias-garcia, line 73: I would remove this log here!

PR #454, opened 2021-04-27
my code title and subtitle changes

mmilenkovic:


src/views/shared/page-header/page-header.view.jsx
```diff
@@ -9,6 +9,7 @@ import { ReactComponent as CloseIcon } from '../../../images/icons/close.svg'
 
 function PageHeader ({
   title,
+  subTitle,
```
elias-garcia, line 12: I think that `subTitle` should be all in lowercase

src/views/shared/page-header/page-header.view.jsx
```diff
@@ -31,7 +32,10 @@ function PageHeader ({
               <ArrowBackIcon />
             </button>
           )}
-          <h1 className={classes.title}>{title || ''}</h1>
+          <div className={classes.titleWrapper}>
+            <h1 className={classes.title}>{title || ''}</h1>
+            <h4 className={classes.subTitle}>{subTitle || ''}</h4>
```
elias-garcia, line 37: Instead of displaying its space if the `subtitle` is not there, I would better do:

```js
{subtitle && <h4 className={classes.subTitle}>{subTitle}</h4>}
```

elias-garcia (approved):

PR #459, opened 2021-04-29
Feature/show current wallet version in my account

mmilenkovic:


src/views/my-account/my-account.view.jsx
```diff
@@ -23,6 +23,9 @@ import Button from '../shared/button/button.view'
 import { copyToClipboard } from '../../utils/browser'
 import { ReactComponent as QRCodeIcon } from '../../images/icons/qr-code.svg'
 
+const packagejson = require('../../../package.json')
```
elias-garcia, line 26: I don't like this `require` here, it would work because `create-react-app` adds polyfills for this but it would be better to change it to `import`

src/views/my-account/my-account.view.jsx
```diff
@@ -150,6 +151,7 @@ const mapDispatchToProps = (dispatch) => ({
       type: 'page',
       data: {
         title: 'My Account',
+        subtitle: packagejsonVersion,
```
elias-garcia, line 154: Maybe this should be now a template string: `version ${packagejsonVersion}`?

mmilenkovic: Why not :)

elias-garcia (approved):

PR #480, opened 2021-05-11
don't CreateAccountAuthorization if receiver is BJJ address

mmilenkovic:


src/views/transaction/components/transaction-form/transaction-form.view.jsx
```diff
@@ -383,7 +383,7 @@ function TransactionForm ({
       case TxType.Transfer: {
         const accountChecks = [
           getAccounts(receiver, [account.token.id]),
-          getCreateAccountAuthorization(receiver).catch(() => {})
+          !isHermezBjjAddress(receiver) ? getCreateAccountAuthorization(receiver).catch(() => {}) : {}
```
elias-garcia, line 386: I think that instead of adding an empty object, not adding anything at all would be more "correct". I propose something like this:

```js
...(  !isHermezBjjAddress(receiver) ? [getCreateAccountAuthorization(receiver).catch(() => {})] : []  )
```

What do you think?

mmilenkovic: Done

elias-garcia (approved):

PR #488, opened 2021-05-17
Shows 0 instead of -- in fiat balance

AlbertoElias:


elias-garcia (approved):

elias-garcia:
Maybe we need to do the same in the `TokenBalance` view, right?

elias-garcia (approved):

PR #493, opened 2021-05-20
Add localStorage v2 migration

elias-garcia:

### DIFF
--- a/src/utils/storage.js
+++ b/src/utils/storage.js
@@ -43,7 +43,6 @@ function runV2Migration () {
   }, {})
 
   setStorage(PENDING_DELAYED_WITHDRAWS_KEY, newPendingDelayedWithdraws)
-  console.log('Storage set up')
 }
 
 function checkVersion () {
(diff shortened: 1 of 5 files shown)


src/utils/storage.js
```diff
@@ -1,4 +1,67 @@
...
+  const currentStorageVersion = JSON.parse(localStorage.getItem(STORAGE_VERSION_KEY))
+
+  if (!currentStorageVersion) {
+    localStorage.setItem(STORAGE_VERSION_KEY, STORAGE_VERSION)
+  }
+
+  // LocalStorage migrations
+  if (currentStorageVersion && currentStorageVersion > STORAGE_VERSION) {
+    // Added L1 withdraws tracking
+    if (STORAGE_VERSION === 2) {
```
AlbertoElias, line 58: >= so that when we change to 3, if someone is still on 1, it would work

src/utils/storage.js
```diff
@@ -1,4 +1,67 @@
...
+
+function checkVersion () {
+  const currentStorageVersion = JSON.parse(localStorage.getItem(STORAGE_VERSION_KEY))
+
+  if (!currentStorageVersion) {
+    localStorage.setItem(STORAGE_VERSION_KEY, STORAGE_VERSION)
+  }
+
+  // LocalStorage migrations
+  if (currentStorageVersion && currentStorageVersion > STORAGE_VERSION) {
```
AlbertoElias, line 56: shouldn't it be when it's lower?

AlbertoElias (approved):

PR #562, opened 2021-06-28
Fix coordinator state loading

elias-garcia:
### What does this PR does?

This PR loads the `coordinator state` when launching the app and then each 30s.

## Checklist

review them:


src/store/global/global.thunks.js
```diff
@@ -86,6 +87,10 @@ function changeRedirectRoute (redirectRoute) {
  */
 function fetchFiatExchangeRates (symbols) {
```
AlbertoElias, line 88: Maybe you don't need the parameter anymore?

src/store/transaction/transaction.thunks.js
```diff
@@ -196,7 +196,9 @@ function fetchAccountBalance () {
  * @returns {void}
  */
 function fetchFees () {
-  return function (dispatch) {
+  return function (dispatch, getState) {
+    const { global: { coordinatorStateTask } } = getState()
+    console.log(coordinatorStateTask)
```
AlbertoElias, line 201: leftover?

AlbertoElias (changes requested):
Couple comments

src/store/global/global.thunks.js
elias-garcia: yeah, nice catch!

AlbertoElias (approved):

PR #565, opened 2021-06-29
Multicoordinator fees

### LINKED ISSUE #510: Multicoordinator fixes for fees + create account authorization
- Uncomment the duplicate create account authorization code + test
- In transaction form, get the correct next forger (the exact next one, or if next + 1 is about to start, next + 1), and fetch its fees.

elias-garcia:
Closes #510.

### What does this PR does?

In this PR we query for the corresponding coordinator API to fetch the fees of a transaction.

## Checklist

review them:

### DIFF
--- a/src/utils/coordinator.js
+++ b/src/utils/coordinator.js
@@ -1,3 +1,5 @@
+import { DELAY_TO_NEXT_FORGER } from '../constants'
+
 /**
  * Extracts the next forgers without duplicates from the coordinator state returned by
  * the Hermez API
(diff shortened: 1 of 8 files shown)


src/utils/coordinator.js
```diff
@@ -0,0 +1,64 @@
...
+ */
+function getNextForgerUrls (coordinatorState) {
+  const nextForgerUrls = getNextForgers(coordinatorState)
+    .map((nextForger) => nextForger.coordinator.URL)
+
+  return nextForgerUrls
+}
+
+/**
+ * Returns the next best forger of the network
```
AlbertoElias, line 33: Would add an explanation of what makes a forger the best next one

src/utils/coordinator.js
```diff
@@ -0,0 +1,64 @@
+import { DELAY_TO_NEXT_FORGER } from '../constants'
+
+/**
+ * Extracts the next forgers without duplicates from the coordinator state returned by
+ * the Hermez API
+ * @param {Object} coordinatorState - Coordinator state returned by the Hermez API
+ * @returns Next forgers
+ */
+function getNextForgers (coordinatorState) {
```
AlbertoElias, line 9: Another way to do this is with a Set, it removes duplicates

elias-garcia: Yep, but it doesn't work with objects, it only takes into account their reference instead of their values to create the set

src/utils/coordinator.js
elias-garcia: done!

AlbertoElias (dismissed):

AlbertoElias (approved):

PR #578, opened 2021-06-30
Remove unused parameter

### LINKED ISSUE #483: Remove redirectTo param
Remove redirectTo param in the tx overview screen

mmilenkovic:
Closes #483 . _in case of a bug fix, this should point to a bug or any other related issue(s)_

### What does this PR does?

Removes unused `redirectTo` parameter from transaction.view page.

### How to test?

Perform different type of transactions.

## Checklist

review them:

### DIFF
--- a/src/views/shared/transaction-info/transaction-info.view.jsx
+++ b/src/views/shared/transaction-info/transaction-info.view.jsx
@@ -87,7 +87,7 @@ function TransactionInfo ({
               value: getPartiallyHiddenHermezAddress(txData.fromHezEthereumAddress)
             }}
             to={{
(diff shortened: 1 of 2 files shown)


mmilenkovic:
Screenshot 2021-06-30 at 18 40 14

@elias-garcia The only place left where this is used is this one ☝️ Do you know if this flow is still valid as is?

elias-garcia:
I think that at the end we agreed with @chamian on redirecting always to a specific page right? So we don't need that param anymore, because it was used to know from which page you enter this view so we can redirect to the same page when you finish the process.

mmilenkovic:
I asked Chami to confirm. Will update the issue accordingly.

mmilenkovic:
Chami:

elias-garcia (approved):

PR #587, opened 2021-07-06
Changes for app not to crash if there is no fromItemHistory

### LINKED ISSUE #554: Wallet crashes when done quickly multiple transfers
## Summary of Bug

The bug happen when transfer multiple times consecutively Usually happens between 5-20 transfers 
The console displays the following error:

```
TypeError: Cannot read property 'reduce' of undefined
    at account-details.thunks.js:164
    at index.js:8
    at dispatch (redux.js:659)
    at account-details.thunks.js:105
    at index.js:8
    at onLoadHistoryTransactions (account-details.view.jsx:362)
    at account-details.view.jsx:114
    at sa (react-dom.production.min.js:211)
    at Bs (react-dom.production.min.js:257)
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
```

I could only reproduce the bug without a selenium script once, usually you have to go pretty fast
### Steps to Reproduce

For example:
- Transfer 50 USDT 10 times consecutively in quickly succession

### Browser information

- Browser: `Chrome`
- Device/OS: `Desktop/Windows 10`

mmilenkovic:
Closes #554 

### What does this PR does?

Add changes for app not to crash if there is no fromItemHistory.

### How to test?

Try quickly going back and forth from one page to another, or perform a number of transactions in short time.

## Checklist

review them:


src/store/home/home.thunks.js
```diff
@@ -143,18 +143,20 @@ function refreshAccounts (hermezEthereumAddress, poolTransactions, pendingDeposi
...
-            hermezEthereumAddress,
-            undefined,
-            fromItem,
-            undefined,
-            undefined,
-            axiosConfig
-          )
-        ]), [initialReq])
+
+      const requests = accountsTask.status === 'reloading' ? [] : accountsTask.data.fromItemHistory
```
elias-garcia, line 147: This doesn't fix the root problem :) The problem looks that when `status` is `reloading`, `fromItemHistory` is not inside the `data` property, and it should.

mmilenkovic: Fixed

elias-garcia (approved):

PR #588, opened 2021-07-06
Replace exchange rates api

elias-garcia:
### What does this PR does?

This PR replaces the exchange rates API.

### How to test?

1. Run the app.
2. Check that all conversions to fiat are working fine.

## Checklist

review them:


src/apis/fiat-exchange-rates.js
```diff
@@ -12,28 +12,15 @@ const baseApiUrl = FIAT_EXCHANGE_RATES_API_URL
...
+      }), {})
+    })
 }
 
-/**
- * Mocks the USD exchange rates for the requested currency symbols
- * @param {string[]} symbols - ISO 4217 currency codes
- * @returns {Object} - USD exchange rates for the requested symbols
- */
-function getMockedFiatExchangeRates () {
```
AlbertoElias, line 26: What if we keep this in case the API fails?

elias-garcia: It could be a great idea but if the API fails for some reason we are not going to detect it, as we are always going to see some values in the screen. I think that is better to fail fast and detect the error as soon as possible. Maybe we can think on a manual switch strategy to use, maybe an env var: `MOCK_EXCHANGE_RATES=true`, so we are able to detect the problem in the API + fix it quickly. I'm not sure... Just throwing out the idea

AlbertoElias: Couldn't we use it in the catch instead of setting everything as undefined like we currently do?

elias-garcia: Yep, we can do it without any problem. The problem is that we are not going to detect when the api is failing as we are always going to see correct values in the fiat conversions... If this is not a problem and we assume that somehow anyone is going to detect this (API tests in the node or whatever) I agree about adding it 👍🏻

AlbertoElias: @elias-garcia Oh, I see what you mean and that's a very good point. Let's bring it up with them, but I do know for sure that they idea is to make the Wallet as user-facing resilient as possible

AlbertoElias: Let's discuss if the option is using fixed values or just showing 0 or -

AlbertoElias (approved):

PR #601, opened 2021-07-13
Refactor amount input

### LINKED ISSUE #569: Token Swap/Reusable functional amount input
Make a reusable component (to also be used in tx form) that can handle all amount value edge cases

elias-garcia:
Closes #569.

### What does this PR does?

This PR refactors the amount input in the transaction form to be reused in similar features like the token swap. Also the code has been refactored entirely to make it simpler and easier to read and make changes in the future.

### How to test?

Try to do a transaction playing around with the amount.

## Checklist

review them:


.env.example
```diff
@@ -12,8 +12,8 @@ REACT_APP_INFURA_API_KEY=example
 # Can be production or development
 REACT_APP_ENV=development
 # Env vars below should only be used when REACT_APP_ENV=development
-REACT_APP_HERMEZ_API_URL=http://localhost:8086
+REACT_APP_HERMEZ_API_URL=https://api.internaltestnet.hermez.io
```
AlbertoElias, line 15: We shouldn't make this public

src/utils/transactions.js
```diff
@@ -40,7 +51,77 @@ function getTxPendingTime (coordinatorState, isL1, timestamp) {
...
+
+        return newMaxAmount.gt(0)
+          ? newMaxAmount
+          : BigNumber.from(0)
+      }
+      default: {
+        const l2FeeBigInt = parseUnits(l2Fee.toFixed(token.decimals), token.decimals)
+        const newMaxAmount = BigNumber.from(getMaxAmountFromMinimumFee(l2FeeBigInt, maxAmount).toString())
+
+        return newMaxAmount.gt(0)
```
AlbertoElias, line 111: I think due to the nature of `getMaxAmountFromMinimumFee`, this value will never be < 0

src/views/transaction/components/fee/fee.view.jsx
```diff
@@ -2,11 +2,11 @@ import React from 'react'
 import { TxType } from '@hermeznetwork/hermezjs/src/enums'
 
 import useFeeStyles from './fee.styles'
-import { MAX_TOKEN_DECIMALS } from '../../../../constants'
-import { CurrencySymbol, getAmountInPreferredCurrency, getTokenAmountInPreferredCurrency } from '../../../../utils/currencies'
+import { CurrencySymbol, getAmountInPreferredCurrency, getFixedTokenAmount, getTokenAmountInPreferredCurrency } from '../../../../utils/currencies'
 import { ReactComponent as AngleDownIcon } from '../../../../images/icons/angle-down.svg'
 import FeesTable from '../fees-table/fees-table.view'
 import { getRealFee } from '../../../../utils/fees'
+import { parseUnits } from 'ethers/lib/utils'
```
AlbertoElias, line 9: Move this to external imports

src/views/shared/amount-input/amount-input.jsx
```diff
@@ -0,0 +1,145 @@
...
+    function handleInputChange (event) {
+      if (INPUT_REGEX.test(event.target.value)) {
+        if (showInFiat) {
+          const newAmountInFiat = Number(event.target.value)
+          const newAmountInTokens = convertAmountToTokens(newAmountInFiat)
+          const fixedAmountInTokens = fixTransactionAmount(newAmountInTokens)
+
+          setAmount({ tokens: fixedAmountInTokens, fiat: newAmountInFiat.toFixed(2) })
+          checkAmountValidity(fixedAmountInTokens)
+          setValue(event.target.value)
```
AlbertoElias, line 84: Why do we need this if we never modify it?

src/views/transaction/components/transaction-amount-input/transaction-amount-input.view.jsx
```diff
@@ -0,0 +1,100 @@
...
+            type='text'
+            onChange={onInputChange}
+          />
+        </div>
+        <div className={classes.amountButtons}>
+          <button
+            type='button'
+            className={`${classes.amountButtonsItem} ${classes.amountButton} ${classes.amountMax}`}
+            tabIndex='-1'
+            // disabled={gasPrice === undefined}
```
AlbertoElias, line 66: can it be removed?

AlbertoElias (changes requested):
Couple small comments. Would also be nice to add comments to the different functions and hooks to make it easier to follow

src/views/shared/amount-input/amount-input.jsx
elias-garcia: As this is a `controlled` input we need to store the value to pass it down to the html `input`. We need this to prevent user from input invalid characters, which is not possible with `uncontrolled` inputs.

AlbertoElias (approved):

PR #619, opened 2021-08-04
Token swap quotes

### LINKED ISSUE #574: Token Swap/Quote Selector
- Double check missing info with Best Offer

elias-garcia:
Closes #574.

### What does this PR does?

This PR adds the `QuoteSelector` and the `QuoteSidenav` components. We still need to figure out how we are handling the icons of the exchanges based on their names. The same remains pending for the icons of the different coins.

## Checklist

review them:


AlbertoElias:
Exchange logos should probably come from the API, can you write it down to ask Oleg?

src/views/token-swap/components/quote-list/quote-list.view.jsx
```diff
@@ -0,0 +1,44 @@
...
+
+  function getBestQuoteAmountDiff (quote) {
+    return BigNumber.from(quote.amountToToken).sub(BigNumber.from(bestQuote.amountToToken))
+  }
+
+  return (
+    <div className={classes.root}>
+      {quotes.map((quote) => (
+        <div
+          key={quote.lpId}
```
AlbertoElias, line 25: Probably worth mentioning somewhere what lp means (liquidity provider?). It's used in several places and will be confusing for someone new

src/views/token-swap/components/quote/quote.view.jsx
```diff
@@ -0,0 +1,67 @@
...
+  toToken,
+  amountToToken,
+  amountToTokenDiff,
+  isTheBest,
+  isSelected,
+  onSelect,
+  onShowMoreInfo
+}) {
+  const classes = useQuoteStyles()
+  console.log(amountToTokenDiff)
```
AlbertoElias, line 19: leftover?

AlbertoElias (changes requested):
Small comments

src/views/token-swap/token-swap.view.jsx
```diff
@@ -10,45 +10,58 @@ import * as tokenSwapActions from '../../store/token-swap/token-swap.actions'
...
   const [amountFrom, setAmountFrom] = React.useState(BigNumber.from(0))
   const [amountTo, setAmountTo] = React.useState(BigNumber.from(0))
   const [selectedTokens, setSelectedTokens] = React.useState({})
-  const [isOfferSidenavOpen, setIsOfferSidenavOpen] = React.useState()
+  const [bestQuote, setBestQuote] = React.useState()
+  const [selectedQuote, setSelectedQuote] = React.useState(undefined)
+
+  React.useEffect(() => {
+    if (quotesTask.status === 'successful') {
+      const bestQuote = quotesTask.data.reduce((quote, bestQuote) => {
```
adrifdez, line 42: Same name that variable from state, could create confusion

src/views/token-swap/token-swap.view.jsx
```diff
@@ -10,45 +10,58 @@ import * as tokenSwapActions from '../../store/token-swap/token-swap.actions'
...
+        if (!bestQuote) {
+          return quote
+        }
+
+        return BigNumber(quote.amountToToken).gt(BigNumber(bestQuote).amountToToken)
```
adrifdez, line 47: I think that you need to invoke BigNumber with new or use from method like `BigNumber.from(quote.amountToToken)`, and the second element compared `BigNumber(bestQuote).amountToToken` should be `BigNumber(bestQuote.amountToToken)`

src/views/token-swap/components/quote/quote.view.jsx
adrifdez, line 19: console.log forgotten

src/views/token-swap/components/quote/quote.view.jsx
```diff
@@ -0,0 +1,67 @@
...
+            More info
+          </button>
+        </div>
+        <div className={classes.toTokenAmountCell}>
+          <p className={classes.toTokenAmount}>
+            {getFixedTokenAmount(amountToToken, toToken.token.decimals)}
+          </p>
+          {amountToTokenDiff.gt(0) && (
+            <p className={classes.toTokenAmountDiff}>
+              -{getFixedTokenAmount(amountToTokenDiff, toToken.token.decimals)}
```
adrifdez, line 46: what happen if I select other quote that is not the best with that -, and the color is always red

adrifdez (changes requested):

src/views/token-swap/components/quote-list/quote-list.view.jsx
elias-garcia: Well it's the unique identifier from the API (yep, it means liquidity provider). But I don't think that it's necessary to explain it somewhere because at the end we are doing the same with a lot of data coming from the Hermez API (for example using `itemId` por the pagination, which is not clear enough without taking a look at the docs)...

src/views/token-swap/components/quote/quote.view.jsx
elias-garcia: The diff is calculated against the `bestQuote`, no matter which quote you've selected. That's why it's going to be always in negative and in red (except in the `bestQuote` row which it shouldn't appear).

src/views/token-swap/token-swap.view.jsx
```diff
@@ -10,45 +10,58 @@ import * as tokenSwapActions from '../../store/token-swap/token-swap.actions'
...
+  const [selectedQuote, setSelectedQuote] = React.useState(undefined)
+
+  React.useEffect(() => {
+    if (quotesTask.status === 'successful') {
+      const newBestQuote = quotesTask.data.reduce((acc, curr) => {
+        if (!acc) {
+          return curr
+        }
+
+        return BigNumber(curr.amountToToken).gt(BigNumber(acc.amountToToken))
```
adrifdez, line 47: This line is still breaking the app should be something like this
`return BigNumber.from(curr.amountToToken).gt(BigNumber.from(acc.amountToToken))`

adrifdez (changes requested):

adrifdez (dismissed):

AlbertoElias (approved):

adrifdez (approved):

PR #621, opened 2021-08-06
Fix some UX problems

### LINKED ISSUE #620: Token Swap/Fix some UX problems
- tokenSwapThunks try to map quotes when is null.
- The initial value of inputs should be empty instead 0.
- Don't call the api when value is zero
- When you change between tokens is not asking for new quotes.
- The dropdown should be closed if you click outside
- Max btn isn't with the correct fontWeight and should be have pointer cursor.
- The list of accounts inside of the dropdown should be have pointer cursor

adrifdez:
Closes #620 

### What does this PR does?

Fix some Ux problems that we have found in the flow of token swap

## Checklist

review them:

### DIFF
--- a/src/apis/token-swap.js
+++ b/src/apis/token-swap.js
@@ -16,11 +16,12 @@ const baseApiUrl = `${process.env.REACT_APP_TOKEN_SWAP_API_URL}`
  * @returns {Array} - Quotes
  */
 function getQuotes (data) {
-  return axios.post(
-      `${baseApiUrl}/quotes`,
-      data,
-      { headers: { 'Content-Type': 'text/plain' } }
-  ).then(res => res.data.quotes)
+  const str = Object.keys(data).reduce(
+    (acc, curr) => `${curr}=${encodeURIComponent(data[curr])}&${acc}`, ''
+  )
+  return axios.get(
+      `${baseApiUrl}/quotes?${str}`
+  ).then(res => res.data)
 }
 
 export {
--- a/src/apis/token-swap.js
+++ b/src/apis/token-swap.js
@@ -4,23 +4,21 @@ const baseApiUrl = `${process.env.REACT_APP_TOKEN_SWAP_API_URL}`
 
 /**
  * Get quotes from liquidity providers
- * @param {Object} data - There should be either amountToToken or
+ * @param {Object} params - There should be either amountToToken or
  * amountFromToken field in request set, not both. In first case client
  * declares the amount of token he wants to receive. In the last case, client
  *  specifies the amount he has for exchange.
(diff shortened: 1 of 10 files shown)


AlbertoElias (dismissed):

src/apis/token-swap.js
```diff
@@ -16,11 +16,12 @@ const baseApiUrl = `${process.env.REACT_APP_TOKEN_SWAP_API_URL}`
  * @returns {Array} - Quotes
  */
 function getQuotes (data) {
-  return axios.post(
-      `${baseApiUrl}/quotes`,
-      data,
-      { headers: { 'Content-Type': 'text/plain' } }
-  ).then(res => res.data.quotes)
+  const str = Object.keys(data).reduce(
```
elias-garcia, line 19: You don't need to build the URI manually with axios. This is enough:

```js
return axios.get(`${baseApiUrl}/quotes`, { params: data })
    .then(res => res.data)
```

src/views/token-swap/components/swap-button/swap-button.view.jsx
```diff
@@ -4,20 +4,19 @@ import PropTypes from 'prop-types'
...
 import useSwapButtonStyles from './swap-button.style'
 
-function SwapButton ({ quotes }) {
+function SwapButton ({
+  quotes,
+  selectedQuote
+}) {
   const classes = useSwapButtonStyles()
-  const [timeUntilValid, setTimeUntilValid] = React.useState(30000)
+  const [timeUntilValid, setTimeUntilValid] = React.useState(selectedQuote?.validUntil.getTime())
```
elias-garcia, line 12: Why do we need this state variable? I think that we are only using it inside the effect to set the `countdown` right?

src/views/token-swap/components/swap-button/swap-button.view.jsx
```diff
@@ -28,30 +27,26 @@ function SwapButton ({ quotes }) {
...
 
-  const buttonLabels = {
-    successful: `Swap ${msToTime(timeUntilValid)}`,
-    failure: 'Insufficient liquidity'
-  }
-
   return (
     <div className={classes.root}>
-      {['successful', 'failure'].includes(quotes.status) &&
+      {selectedQuote &&
```
elias-garcia, line 32: I think that this should be `countdown && ...` as we are now using countdown below instead the `selectedQuote` right?

elias-garcia (changes requested):

src/views/token-swap/components/swap-button/swap-button.view.jsx
adrifdez: selectQuote covers more cases

elias-garcia (approved):

PR #624, opened 2021-08-11
Withdraws with snark

### LINKED ISSUE #121: Withdraws with Snark
**Move from withdrawMerkleProof to withdrawCircuit.**

The objective of this task is to replace the current `withdraw` with the new `withdrawCircuit`, which is cheaper in terms of fee when more than one exit is present in a batch.

In the Wallet, we currently call the #withdraw function in hermezJS: 
  * https://github.com/hermeznetwork/wallet-ui/blob/main/src/store/transaction/transaction.thunks.js#L284

But now we need to use #withdrawCircuit 
  * https://github.com/hermeznetwork/hermezjs/blob/main/src/tx.js#L227

Withdraws are a 2 step process:

  * **Exit** in HZ smart contract
    * The tx is forged and the funds are blocked
    * A single batch can have multiple exits (one per account)
  * **Final Withdraw** Move funds from L2 to L1

More details on this are in the [docs](https://docs.hermez.io/#/developers/sdk?id=withdrawing).

Notes:
  * Withdraw identifiers are based on both the batch and the account
  * Currently, we only support 1 "exit" per account and batch.
  * Multiple accounts can be linked to a single BJJ (user public key)
  * Check the tests available in the library for the `withdrawCircuit`
  * All this happens in the transaction overview and thunks
 
Type of accounts:
  * **Internal account**: An account linked only to a BJJ
  * **Normal account**: An account linked a BJJ and to an ethereum address

amonsosanz:
Closes #121.

### What does this PR does?

This PR replaces the current `withdraw` with the new `withdrawCircuit`, which is cheaper in terms of fee when more than one exit is present in a batch.

## Checklist

review them:


package.json
```diff
@@ -58,7 +71,11 @@
     "production": [
       ">0.2%",
       "not dead",
-      "not op_mini all"
+      "not op_mini all",
+      "not ie <=11",
+      "not android <10",
+      "not safari <14",
+      "not ios_saf <14"
```
AlbertoElias, line 78: Can you also change the `development` ones to make them the same

src/serviceWorkerRegistration.js
```diff
@@ -8,18 +8,50 @@
...
 
 const isLocalhost = Boolean(
   window.location.hostname === 'localhost' ||
     // [::1] is the IPv6 localhost address.
     window.location.hostname === '[::1]' ||
-    // 127.0.0.1/8 is considered localhost for IPv4.
-    window.location.hostname.match(
-      /^127(?:\.(?:25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)){3}$/
-    )
+    // 127.0.0.0/8 are considered localhost for IPv4.
```
AlbertoElias, line 17: is considered*

package.json
```diff
@@ -29,11 +29,24 @@
...
     "react-redux": "^7.2.0",
     "react-router-dom": "^5.2.0",
-    "react-scripts": "^4.0.2",
+    "react-scripts": "^4.0.3",
     "redux": "^4.0.5",
     "redux-devtools-extension": "^2.13.8",
     "redux-thunk": "^2.3.0",
-    "serve": "^11.3.2"
+    "serve": "^11.3.2",
+    "workbox-background-sync": "^5.1.3",
```
AlbertoElias, line 37: Why so many? It seems like you use 3 in the service worker

src/store/transaction/transaction.thunks.js
```diff
@@ -292,14 +292,11 @@ function withdraw (amount, account, exit, completeDelayedWithdrawal, instantWith
...
-        amount,
-        account.accountIndex,
-        account.token,
-        wallet.publicKeyCompressedHex,
-        exit.batchNum,
-        exit.merkleProof.siblings,
+      return Tx.withdrawCircuit(
+        exit,
         instantWithdrawal,
+        'https://raw.githubusercontent.com/hermeznetwork/hermezjs/main/withdraw-circuit-files/withdraw.wasm',
```
AlbertoElias, line 298: Maybe they can be constants?

AlbertoElias (changes requested):
Just a couple comments! :)

package.json
amonsosanz: Sure. Done!

src/serviceWorkerRegistration.js
amonsosanz: It's funny because I didn't write this myself. Fixed.

src/store/transaction/transaction.thunks.js
amonsosanz: I thought about that especially because they are used in 2 modules. Would you declare them as global constants somewhere and import them here and in `service-worker.js` or do we do local declarations?

package.json
AlbertoElias: I think it's better to minimize. npm i can get very long, and it also makes the `package.json` easier to maintain. I think CRA is a great starting point, but as an application evolves and grows, it's good to change things to adapt it.

src/store/transaction/transaction.thunks.js
AlbertoElias: I would use the constants file that we have

package.json
elias-garcia: Yep, I was about to comment the same about this dependencies. I would also add only those which are necessary.

src/store/transaction/transaction.thunks.js
elias-garcia: Same here, we have a `constants.js` file under the `src` folder to keep this things easy to change and shareable between different modules.

elias-garcia (commented):
Can we also write the `serviceWorkerRegistration.js` filename in kebab case instead of camel case as it's the standard we are following in the repo?

package.json
amonsosanz: Done.

src/store/transaction/transaction.thunks.js
amonsosanz: Importing the URLs is not working on the service worker. I've been reading about this and it looks like service workers do not fully support ES6 imports. So far I've moved the URLs to constants and used them on `transaction.thunks.js` but not on `service-worker.js`. I added a comment in constants to make sure we do update the URL's in the service worker when we update them in the constants.

Maybe you have more experience with service workers and can point to a solution that allows us to import the constants there?

AlbertoElias (approved):

PR #626, opened 2021-08-17
Allow reload when some calls fail

### LINKED ISSUE #522: Reloading stops making requests if one fails
We have an auto refresh that makes requests to get the latest info. If one of the requests in auto refresh fails, specifically /exits in home, it stops retrying

adrifdez:
Closes #522.

### What does this PR does?

Remove the logic that was blocking reload if the previous attempt failed

## Checklist

review them:


AlbertoElias (approved):
I am approving this but pinging @elias-garcia to take a look for when he comes back

PR #639, opened 2021-09-01
Priceupdater

### LINKED ISSUE #618: Move to new price updater API
https://github.com/hermeznetwork/price-updater-service

[] Move fiat conversion API to this. This should just be changing the constant for fiat prices.
[] Move all token values from price updater. This requires that all token information currently fetched from the Coordinator's API should now be fetched from this new API. Ask Tiago for info on where testnet deployments are.

adrifdez:
Closes #618
### What does this PR does?

Important, This is a draft to typescript branch to receive feedback meanwhile that branch merges to develop.

Now all the values related to fiat are coming from a new source "price-updater". So we replace the value coming from coordinator with new one

## Checklist

review them:


AlbertoElias:
Couple things:

- Why is the package-lock updated?
- Why is it adding semicolons everywhere?

src/domain/hermez/index.ts
```diff
@@ -7,8 +7,8 @@
  * them as required and finally remove all those not required.
  */
 
-import { BigNumber } from 'ethers'
-import { ISOStringDate } from '../'
+import { BigNumber } from "ethers";
```
AlbertoElias, line 10: Why double quotes instead of single quotes?

adrifdez:
> Couple things:
> 
> * Why is the package-lock updated?
> * Why is it adding semicolons everywhere?
> * Why is it changing to double quotes?
> 
> I don't think we should apply any code style changes. It also makes it harder to review as all those diffs are mixed in

package-lock was updated from an old version of npm and this one recover the last format 

The new ts files have been formatted  with normal prettier rules so it's changing to double quotes and adding semicolons

src/apis/price-updater.ts
```diff
@@ -0,0 +1,58 @@
+import axios from "axios";
+
+import { CurrencySymbol } from "../utils/currencies";
+
+// domain
+import { FiatExchangeRates, Tokens } from "../domain/hermez";
+
+interface ApiExchangeRate {
```
elias-garcia, line 8: This should go in the persistence level, at the end this is information coming from the outside that we need to convert to our domain (which you are already doing perfectly later).

src/store/global/global.thunks.ts
```diff
@@ -14,9 +14,9 @@ import { HttpStatusCode } from "@hermeznetwork/hermezjs/src/http";
 import { getEthereumAddress } from "@hermeznetwork/hermezjs/src/addresses";
 
 import * as globalActions from "./global.actions";
-import * as fiatExchangeRatesApi from "../../apis/fiat-exchange-rates";
 import * as hermezWebApi from "../../apis/hermez-web";
 import * as airdropApi from "../../apis/rewards";
+import * as priceApi from "../../apis/price-updater";
```
elias-garcia, line 19: `import * as priceUpdaterApi from "../../apis/price-updater";`

src/views/shared/amount-input/amount-input.view.jsx
```diff
@@ -101,6 +99,7 @@ function AmountInput (Component) {
      * @param {InputEvent} event - Input event
      */
     function handleInputChange (event) {
+      const INPUT_REGEX = new RegExp(`^\\d*(?:\\.\\d{0,${account?.token.decimals}})?$`)
```
elias-garcia, line 102: Why moving this regex to the input change? It's better to have it as a constant at the top as it was before to prevent creating it at every input change. Also the `?` is not needed as when we create the account we are going to add the token.

src/store/token-swap/token-swap.thunks.js
```diff
@@ -18,7 +14,8 @@ function fetchAccounts (fromItem) {
     const {
       global: {
         wallet: { publicKeyBase64: hermezAddress },
-        fiatExchangeRatesTask: { data: fiatExchangeRates }
+        fiatExchangeRatesTask: { data: fiatExchangeRates },
```
elias-garcia, line 17: Here you are assuming that the `fiatExchangeRatesTask` is succesful which is not true. Aren't you getting a TypeScript warning here?

src/store/account-details/account-details.thunks.ts
```diff
@@ -1,102 +1,132 @@
...
+    dispatch(accountDetailsActions.loadAccount());
 
     return CoordinatorAPI.getAccount(accountIndex)
       .then((account: Account) => {
         if (wallet === undefined || account.bjj !== wallet.publicKeyBase64) {
-          dispatch(push('/'))
+          dispatch(push("/"));
         } else {
-          dispatch(accountDetailsActions.loadAccountSuccess(account))
+          const accountTokenUpdated =
```
elias-garcia, line 35: I would create another `createAccount` function which receives an `Account` from the persistence layer and the `pricesTask` and returns the new `Account` from the domain.

elias-garcia (changes requested):

src/views/shared/amount-input/amount-input.view.jsx
adrifdez: The safe call "?" cover a bug with token swap because render it without know an account, and we moved it inside because was a bug in the matic swapper that was copied from this example

src/store/global/global.actions.ts
```diff
@@ -65,6 +66,9 @@ export enum GlobalActionTypes {
   LOAD_REWARD_TOKEN = "[GLOBAL] LOAD REWARD TOKEN",
   LOAD_REWARD_TOKEN_SUCCESS = "[GLOBAL] LOAD REWARD TOKEN SUCCESS",
   LOAD_REWARD_TOKEN_FAILURE = "[GLOBAL] LOAD REWARD TOKEN FAILURE",
+  LOAD_TOKENS_PRICE = "[GLOBAL] LOAD TOKENS PRICE",
```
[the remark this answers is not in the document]
adrifdez: it's a list of tokens with one price, so still think the same way?

src/store/token-swap/token-swap.thunks.js
adrifdez: True, we will need to improve that flow if we continue with token swap but for now `getTokenAmountInPreferredCurrency` has a conditional to check data value

src/views/shared/amount-input/amount-input.view.jsx
elias-garcia: Which was the bug of having the regex outside this function?

src/store/global/global.actions.ts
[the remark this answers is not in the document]
elias-garcia: Yep, because at the end it's a list of token prices right?.

src/store/token-swap/token-swap.thunks.js
elias-garcia: But what we can do by the moment is not using object destructuring, and doing just: `fiatExchangeRatesTask,` and check that it's `successful` below.

src/views/shared/amount-input/amount-input.view.jsx
adrifdez: The rule needs know the maximum decimals of the token to not break the bignumber format

elias-garcia (approved):

PR #650, opened 2021-09-09
Notifications on transactions status

### LINKED ISSUE #390: Improvement of flow for transaction confirmation and error
### **Confirmation**

L2 transfers: After clicking "Send" (https://zpl.io/2EMD44v), show notification of "Transaction Submitted" in Token account Activity.
https://zpl.io/2p85PGr

Deposit and Withdraw: After signing with Metamask, show notification of "Transaction Submitted" in the Homepage.
https://zpl.io/V0vKp89
https://zpl.io/VxAErAW

_______

### **Error**

**Instant error after confirming transaction:**

L2 transfers: Show error notification in the Token Account activity + "Invalid" label in the transaction.
https://zpl.io/297rz4p

Deposit and withdraw: Show error notification in the Homepage (black card doesn't appear).
https://zpl.io/VYL31wk

**Error after Pending status of transaction:**

L2 transfers:  In Token Account activity, only show "Invalid" label under the transaction

Deposit and withdraw: "Transaction failed" notification pops up in the Homepage and black card disappears.
https://zpl.io/VYL31wk

adrifdez:
Closes #390

### What does this PR does?

Show a notification with the status of the last transaction and change a small improvement on the flow of token swap when load quotes

image
image

## Checklist

review them:

### DIFF
--- a/src/store/transaction/transaction.thunks.js
+++ b/src/store/transaction/transaction.thunks.js
@@ -309,7 +309,7 @@ function deposit(amount, account) {
       global: { wallet, signer },
     } = getState();
 
-    dispatch(transactionActions.startTransactionSigning());
+    dispatch(transactionActions.startTransactionApproval());
 
     return Tx.deposit(
       HermezCompressedAmount.compressAmount(amount),
@@ -338,8 +338,8 @@ function deposit(amount, account) {
       })
       .catch((error) => {
         console.error(error);
-        dispatch(transactionActions.stopTransactionSigning());
-        handleTransactionFailed(dispatch, error);
+        dispatch(transactionActions.stopTransactionApproval());
+        handleTransactionFailure(dispatch, error);
       });
   };
 }
@@ -351,7 +351,7 @@ function withdraw(amount, account, exit, completeDelayedWithdrawal, instantWithd
     } = getState();
     const withdrawalId = account.accountIndex + exit.batchNum;
 
-    dispatch(transactionActions.startTransactionSigning());
+    dispatch(transactionActions.startTransactionApproval());
 
     // Differentiate between a withdraw on the Hermez SC and the DelayedWithdrawal SC
     if (!completeDelayedWithdrawal) {
@@ -396,8 +396,8 @@ function withdraw(amount, account, exit, completeDelayedWithdrawal, instantWithd
         })
         .catch((error) => {
           console.error(error);
-          dispatch(transactionActions.stopTransactionSigning());
-          handleTransactionFailed(dispatch, error);
+          dispatch(transactionActions.stopTransactionApproval());
+          handleTransactionFailure(dispatch, error);
         });
     } else {
       return Tx.delayedWithdraw(wallet.hermezEthereumAddress, account.token, signer)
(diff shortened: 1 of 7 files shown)


src/views/transaction/transaction.view.jsx
```diff
@@ -141,6 +143,18 @@ function Transaction({
 
   React.useEffect(() => onCleanup, [onCleanup]);
 
+  function finishTransaction() {
```
elias-garcia, line 146: We can't do this here. The idea of this task is to remove the `TransactionConfirmation` and `TransactionError` components. This code should be inside the functions that are currently executing the asynchronous tasks: `deposit`, `transfer`, `withdraw`, `forceExit` and `exit` in `transaction.thunks.js`. The idea is also to display the error next to the `Transaction failed` message.

adrifdez: I've moved logic from view to thunks

src/store/transaction/transaction.thunks.js
```diff
@@ -462,6 +452,9 @@ function exit(amount, account, fee) {
     const {
       global: { wallet, coordinatorStateTask },
     } = getState();
+
+    dispatch(transactionActions.startTransactionSigning());
```
AlbertoElias, line 456: Signing? An exit is an L2 tx

src/store/transaction/transaction.thunks.js
```diff
@@ -485,6 +478,9 @@ function transfer(amount, from, to, fee) {
...
```
AlbertoElias, line 482: This is also an L2 tx, and signing is done with Metamask for L1 txs

src/store/transaction/transaction.thunks.js
```diff
@@ -494,15 +490,25 @@ function transfer(amount, from, to, fee) {
...
   };
 }
 
+function handleTransactionSuccess(dispatch, accountIndex) {
+  const route = accountIndex ? `/accounts/${accountIndex}` : "/";
+  dispatch(openSnackbar("Transaction submitted"));
+  dispatch(push(route));
+}
+
+function handleTransactionFailed(dispatch, error) {
```
AlbertoElias, line 508: rename Failed to Failure so it's a noun like Success

AlbertoElias (commented):
Leaving a couple comments

src/store/transaction/transaction.thunks.js
adrifdez: I'm reusing that function to be aware in the view if the transaction fail and reactivate the btn. Maybe I can rename that action to startTransaction and endTransaction

src/store/transaction/transaction.thunks.js
adrifdez: I'm reusing that function to be aware in the view if the transaction fail and reactivate the btn. Maybe I can rename that action to startTransaction and endTransaction

AlbertoElias: I would suggest something like that, yes. Let's see what @elias-garcia thinks

elias-garcia: I agree with @AlbertoElias here, maybe `startTxApproval`? What do you think guys, I think that it's more generic than `startTxSigning`?

adrifdez: I'm still thinking that maybe `startTransaction` is more generic than `startTxApproval` because in L2 you don't need more user steps

AlbertoElias: Thing is, when you start a transaction might be considered the transaction form. I think Approval is a good generic word.

AlbertoElias (approved):

PR #654, opened 2021-09-15
Fix duplicated tx

### LINKED ISSUE #648: Repeated TX are shown in the historic TX list
I have 61 historical TX in my HEZ account (accounts/hez:HEZ:5468) that are obtained in 4 batches (20, 20, 20, 1) using infinite scrolling.

If I leave the page opened though, it keeps polling `/transactions-history` and keeps adding the same transactions to the list, again and again.

Also, there's a duplicated key error as a result of the bug:

image

adrifdez:
Closes #648 

### What does this PR does?

Reloaded history was adding same tx to the list so increasing the list every time that the list is refreshed, now we manage to update the status and check if we need to update the list

## Checklist

review them:

### DIFF
--- a/src/store/account-details/account-details.thunks.ts
+++ b/src/store/account-details/account-details.thunks.ts
@@ -128,7 +128,10 @@ function fetchHistoryTransactions(
       accountDetails: { historyTransactionsTask },
     } = getState();
 
-    if (fromItem === undefined && historyTransactionsTask.status === "successful") {
+    if (
+      fromItem === undefined &&
+      ["reloading", "successful"].includes(historyTransactionsTask.status)
+    ) {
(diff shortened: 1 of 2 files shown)


src/store/account-details/account-details.thunks.ts
```diff
@@ -128,7 +128,10 @@ function fetchHistoryTransactions(
       accountDetails: { historyTransactionsTask },
     } = getState();
 
-    if (fromItem === undefined && historyTransactionsTask.status === "successful") {
+    if (
+      fromItem === undefined &&
+      ["reloading", "successful"].includes(historyTransactionsTask.status)
```
AlbertoElias, line 133: I would keep using the other pattern just for consistency and readability even if this is more clever

AlbertoElias (approved):

PR #655, opened 2021-09-15
check toHezEthereumAddress

### LINKED ISSUE #651: Bug TX Transfer
When we access to see details of a sent transfer dated on current day, it's breaking the app

adrifdez:
Closes #651 

### What does this PR does?

Fix the problem on transactions details that crash the app when the returned data comes without data on the destination bjj or hezEthereumAddres 

## Checklist

review them:

### DIFF
--- a/src/views/shared/transaction-info/transaction-info.view.jsx
+++ b/src/views/shared/transaction-info/transaction-info.view.jsx
@@ -91,13 +91,15 @@ function TransactionInfo({
               subtitle: "My Hermez address",
               value: getPartiallyHiddenHermezAddress(txData.fromHezEthereumAddress),
             }}
-            to={{
-              subtitle:
-                txData.toHezEthereumAddress?.toLowerCase() ===
-                INTERNAL_ACCOUNT_ETH_ADDR.toLowerCase()
-                  ? getPartiallyHiddenHermezAddress(txData.toBjj || txData.toBJJ)
-                  : getPartiallyHiddenHermezAddress(txData.toHezEthereumAddress),
-            }}
+            to={
+              txData.toHezEthereumAddress && {
+                subtitle:
+                  txData.toHezEthereumAddress.toLowerCase() ===
+                  INTERNAL_ACCOUNT_ETH_ADDR.toLowerCase()
+                    ? getPartiallyHiddenHermezAddress(txData.toBjj || txData.toBJJ)
+                    : getPartiallyHiddenHermezAddress(txData.toHezEthereumAddress),
+              }
+            }
             date={date}
             fee={txData.fee}
             preferredCurrency={preferredCurrency}
--- a/src/views/shared/transaction-info/transaction-info.view.jsx
+++ b/src/views/shared/transaction-info/transaction-info.view.jsx
@@ -93,14 +93,14 @@ function TransactionInfo({
             status={getTransactionStatus()}
             from={myHermezAddress}
             to={
-              (txData.toHezEthereumAddress && {
+              (txData.toBJJ && {
                 subtitle:
                   txData.toHezEthereumAddress.toLowerCase() ===
                   INTERNAL_ACCOUNT_ETH_ADDR.toLowerCase()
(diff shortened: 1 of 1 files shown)


src/views/shared/transaction-info/transaction-info.view.jsx
```diff
@@ -91,13 +91,15 @@ function TransactionInfo({
...
             }}
-            to={{
-              subtitle:
-                txData.toHezEthereumAddress?.toLowerCase() ===
-                INTERNAL_ACCOUNT_ETH_ADDR.toLowerCase()
-                  ? getPartiallyHiddenHermezAddress(txData.toBjj || txData.toBJJ)
-                  : getPartiallyHiddenHermezAddress(txData.toHezEthereumAddress),
-            }}
+            to={
+              txData.toHezEthereumAddress && {
```
AlbertoElias, line 95: And what if it's a tx to an internal account and it has a BJJ but no ethereum address?

adrifdez: The last code wasn't covering that scenario so I assume that it's rare to happen. This fix is covering a corner case where those values aren't coming. In theory should be return at least the value that was submitted.
At least with this fix doesn't crash the app in your case either

AlbertoElias: But with this change, internal transactions will stop showing the To field, and we can't allow that.

adrifdez: we can show the same address if it's the known case and cover for the unexpected

src/views/shared/transaction-info/transaction-info.view.jsx
```diff
@@ -42,6 +43,10 @@ function TransactionInfo({
     if (txData.state === TxState.Forged) {
       return { subtitle: TxStatus.Confirmed };
     }
+    console.log(txData);
```
AlbertoElias, line 46: Leftover log

src/views/shared/transaction-info/transaction-info.view.jsx
```diff
@@ -42,6 +43,10 @@ function TransactionInfo({
...
+    if (txData.errorCode) {
```
AlbertoElias, line 47: This should probably go in a different PR

src/views/shared/transaction-info/transaction-info.view.jsx
```diff
@@ -84,20 +89,24 @@ function TransactionInfo({
...
+            from={myHermezAddress}
+            to={
+              (txData.toHezEthereumAddress && {
+                subtitle:
+                  txData.toHezEthereumAddress.toLowerCase() ===
+                  INTERNAL_ACCOUNT_ETH_ADDR.toLowerCase()
+                    ? getPartiallyHiddenHermezAddress(txData.toBjj || txData.toBJJ)
+                    : getPartiallyHiddenHermezAddress(txData.toHezEthereumAddress),
+              }) ||
+              (txData.fromAccountIndex == txData.toAccountIndex && myHermezAddress)
```
AlbertoElias, line 108: ===?

src/views/shared/transaction-info/transaction-info.view.jsx
```diff
@@ -84,20 +89,24 @@ function TransactionInfo({
...
-            to={{
-              subtitle:
-                txData.toHezEthereumAddress?.toLowerCase() ===
-                INTERNAL_ACCOUNT_ETH_ADDR.toLowerCase()
-                  ? getPartiallyHiddenHermezAddress(txData.toBjj || txData.toBJJ)
-                  : getPartiallyHiddenHermezAddress(txData.toHezEthereumAddress),
-            }}
+            from={myHermezAddress}
+            to={
+              (txData.toHezEthereumAddress && {
```
AlbertoElias, line 101: If we can, let's change this to `toBjj`

src/views/shared/transaction-info/transaction-info.view.jsx
adrifdez: yes

AlbertoElias (approved):

PR #657, opened 2021-09-16
Refactor to typescript slice 03 (Fix some linter errors and refactor local storage persistence)

### LINKED ISSUE #653: Refactor to Typescript (slice 03)
This slice is mainly focused on:

  * Fixing some of the linting errors
  * Decoding the data we get from the local storage and some APIs.
  * Disable the linter for the paths that are not yet migrated to TS.
  * Refactor the local storage to be type-safe

amonsosanz:
Closes #653.

### What does this PR does?

This slice is mainly focused on:

  * Fixing all the current linting errors
  * Decoding the data we get from the local storage and some APIs.
  * Disable the linter for the paths that are not yet migrated to TS.
  * Refactor the local storage to be type-safe

## Checklist

review them:


adrifdez (approved):

src/hermezjs-typings.d.ts
```diff
@@ -191,34 +190,34 @@ declare module "@hermeznetwork/*" {
...
-  //   slotNum: number;
-  //   fromBlock: number;
-  //   toBlock: number;
-  //   fromTimestamp: string;
-  //   toTimestamp: string;
-  // }
+  interface Period {
+    // slotNum: number;
+    // fromBlock: number;
+    // toBlock: number;
```
AlbertoElias, line 203: Why all these commented properties? Isn't it better to just remove them?

src/hermezjs-typings.d.ts
```diff
@@ -1,5 +1,6 @@
+/* eslint-disable no-empty */
```
AlbertoElias, line 1: Why this rule?

src/store/account-details/account-details.thunks.ts
```diff
@@ -225,7 +241,10 @@ function refreshHistoryTransactions(accountIndex: Account["accountIndex"], histo
         })
         .then((historyTransactions: Transactions) =>
           dispatch(accountDetailsActions.refreshHistoryTransactionsSuccess(historyTransactions))
-        );
+        )
+        // ToDo: Shouldn't we dispatch refreshHistoryTransactionsFailure here?
```
AlbertoElias, line 245: I think it's better to catch and ignore and just fail silently and try again in the next refresh. No need to show anything in the UI

src/utils/ethereum.ts
```diff
@@ -34,32 +48,37 @@ async function getTokens(wallet, hermezTokens) {
...
-          contract
-            .balanceOf(ethereumAddress)
-            // We can ignore if a call to the contract of a specific token fails.
-            .catch(() => {})
-        );
+        // We can ignore if a call to the contract of a specific token fails.
+        // ToDo: Find a way to properly type the functions of the contract declared through the ABI.
+        // eslint-disable-next-line
+        const balance: Promise<BigNumber> = contract
+          .balanceOf(ethereumAddress)
```
AlbertoElias, line 63: doesn't balanceOf already return a BigNumber?

src/utils/ethereum.ts
```diff
@@ -69,33 +88,40 @@ async function getTokens(wallet, hermezTokens) {
...
  * @returns {Boolean}
  */
-function isTxExpectedToFail(tx, date, accountEthBalance) {
-  if (tx?.blockNumber === null) {
+function isTxExpectedToFail(
+  tx: TransactionResponse,
+  date: ISOStringDate,
+  accountEthBalance: BigNumber
+): boolean {
+  // ToDo: According to ethers types tx?.blockNumber can not be null
```
AlbertoElias, line 122: Might be worth testing to confirm. I do seem to remember it returning null though. Best way is to force a tx to fail by setting a very low gas (21000)

AlbertoElias (commented):
Leaving a few comments and questions

src/apis/price-updater.ts
```diff
@@ -1,20 +1,43 @@
 import axios from "axios";
+import * as z from "zod";
 
 import { CurrencySymbol } from "src/utils/currencies";
 
 // domain
 import { FiatExchangeRates, Token } from "src/domain/hermez";
 
+// persistence
```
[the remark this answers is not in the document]
amonsosanz: Do you mean to remove the comments and the empty lines or just the empty lines?

src/hermezjs-typings.d.ts
amonsosanz: I thought of adding all the props present on the entities of the lib and leave commented and not implemented those that are not currently used by the wallet. On each PR, I uncomment some new props that become used by the new modules I refactor. The idea is that when we end the refactor we will have a clear picture of what props do we use and what don't.

At that point, we could opt to remove the commented features or implement them all since, at the end of the day, these types definition module should not belong to the app, but to the lib. This module describes the types of the library and I think that eventually should cover all the features of the library, no matter whether they are used on the wallet on not. 

Since it's a library, I think some other projects using it may require some other stuff. Does it make sense?

src/hermezjs-typings.d.ts
amonsosanz: We want this rule enabled because we want to disallow empty block statements in normal modules. This though, is a type definition module, and here, exceptionally, we need empty blocks, since the syntax to declare a function's signature is as follows:

```
declare function foo(bar: string): string {};
```

If we don't globally disable the rule in this module the linter would complain about each empty block: `{}`.

src/store/account-details/account-details.thunks.ts
amonsosanz: Ok. Makes sense. I've removed the comment.

src/utils/ethereum.ts
amonsosanz: Don't blame me cause I may be completely lost here... Isn't `balanceOf` a feature of the smart contract? It's described in the `ABI` as a `uint256` so I assumed it was an unsigned integer. This is at least what I understand from the [docs](https://docs.soliditylang.org/en/develop/types.html#integers).

src/utils/ethereum.ts
amonsosanz: I feel this is difficult to test. I mean, I can confirm that it's not `null` in my tests, but how can we guarantee that it will not be `null` in the future? Unless you guys know it can be null for some strange reason, I would stick to the definition of the type from `ethers`, where the prop is optional, so its value can only be `number` or `undefined`:

image

src/apis/price-updater.ts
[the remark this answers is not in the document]
AlbertoElias: Jus the empty lines, I love comments haha

src/hermezjs-typings.d.ts
AlbertoElias: Makes sense. Probably worth mentioning these decisions in the PR itself in the future

src/hermezjs-typings.d.ts
AlbertoElias: Should we consider disabling this globally as we use this a lot including in catches? @elias-garcia

src/utils/ethereum.ts
AlbertoElias: uint256 is a type in Solidity, and then in JS it's a BigInt, but ethers wraps it in a BigNumber

src/utils/ethereum.ts
AlbertoElias: I think @elias-garcia wrote this so he might remember better. I vaguely remember it being null before a tx has been processed

src/hermezjs-typings.d.ts
```diff
@@ -136,12 +140,47 @@ declare module "@hermeznetwork/*" {
...
+    // requestFee?: unknown;
+    // requestFromAccountIndex?: unknown;
+    // requestNonce?: unknown;
+    // requestToAccountIndex?: unknown;
+    // requestToBJJ?: unknown;
+    // requestToHezEthereumAddress?: unknown;
+    // requestTokenId?: unknown;
+  };
+
+  // ToDo: According to the docs: https://apidoc.hermez.network/#model-TransactionType supported types are:
```
AlbertoElias, line 156: What is the ToDo? What is missing?

src/apis/price-updater.ts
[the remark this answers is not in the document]
amonsosanz: I've investigated and concluded that there's no way to enforce this with either prettier or an eslint rule, so it's difficult to maintain consistency in the future.

For now, I've manually removed all the spaces in this [commit](https://github.com/hermeznetwork/wallet-ui/pull/663/commits/ee3c0d8b963dd6029182b3b00bc107a82076a5a3), but I was in the branch of slice 4 by mistake. Since it's not an urgent change, is it ok to leave the change for the PR of the slice 4?

src/hermezjs-typings.d.ts
amonsosanz: Yeah, I've had to disable it per line in several catches. I'm ok disabling it  globally, not a big deal.

src/utils/ethereum.ts
amonsosanz: Yeah that makes sense. Line removed.

src/hermezjs-typings.d.ts
amonsosanz: We miss resolving the inconsistency between the types declared in the docs and the types present in the code:

image

src/apis/price-updater.ts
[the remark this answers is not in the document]
AlbertoElias: Sure, let's leave it like this. Yup, this is hard to enforce, but if prettier also doesn't enforce a new line between imports, then we can use PRs to maintain the coding style as well

src/hermezjs-typings.d.ts
AlbertoElias: In general, the protocol has way more features than the ones we need. We also add a type just for simplicity internally. I think it's good as is. I don't expect us to support the missing types. Definitely not in the mid-term

amonsosanz: When you refer to "the ones *we* need" who do you mean by "we"? "We" as the creators of the `wallet-ui` frontend or "we" as the creators of the `hermezjs` library? I ask because the module `hermezjs-typings.d.ts` should implement any feature available in the library. In other words:

 * Are we ok not supporting in the library the types `CreateAccountDepositTransfer`, `DepositTransfer` and `ForceTransfer` present in the docs  because any user of the lib would need these types implemented by the protocol?
* Is the type `Withdrawn` we are adding generic enough to be used by any client of the library? Or is it just something required by our `wallet-ui` that would not make sense to other users of the lib?

src/apis/hermez-web.ts
```diff
@@ -1,17 +1,31 @@
...
 
 import { HERMEZ_WEB_URL } from "src/constants";
 
 const baseApiUrl = HERMEZ_WEB_URL;
 
+interface WalletUnderMaintenanceResponse {
+  isWalletUnderMaintenance: number;
+}
+
+const walletUnderMaintenanceResponseParser: z.ZodSchema<WalletUnderMaintenanceResponse> = z.object({
```
elias-garcia, line 12: Shouldn't we move this `walletUnderMaintenanceResponseParser` to the persistence module and `WalletUnderMaintenanceResponse` to the domain?

src/hermezjs-typings.d.ts
elias-garcia: Well, in my opinion ESLint rules are to comply with them, not to avoid them in specific lines or files. In this case the `no-empty` rules prevents from having empty blocks of code, like empty `if {}` etc. So in my opinion this rule should stay as it is and we need to delete this comment. If we cannot remove it now because of the refactor we need to do it later.

src/views/account-details/account-details.view.jsx
```diff
@@ -359,7 +361,7 @@ const mapStateToProps = (state) => ({
 });
 
 const mapDispatchToProps = (dispatch) => ({
-  onLoadAccount: (accountIndex) => dispatch(accountDetailsThunks.fetchAccount(accountIndex)),
+  onLoadAccount: (accountIndex, fiatExchangeRates, preferredCurrency) => dispatch(accountDetailsThunks.fetchAccount(accountIndex, fiatExchangeRates, preferredCurrency)),
```
elias-garcia, line 364: Why sending all this extra data instead of picking them from `redux` in the thunks? I think that this is makes views more verbose.

src/utils/ethereum.ts
elias-garcia: I don't remember exactly if it's null or undefined... We would need to test it. This use case is when you put a low gas fee and the transaction hasn't been mined yet. If `ethers` says that is undefined it should be, as they use TypeScript in their codebase and if not they should have compilation errors hehe

src/utils/ethereum.ts
```diff
@@ -1,16 +1,30 @@
...
+function getTokens(
+  wallet: Wallet,
+  hermezTokens: Token[]
+): Promise<
+  {
+    balance: BigNumber;
+    token: Token;
+  }[]
+> {
+  // ToDo: Can we remove this check below?
```
elias-garcia, line 27: If you receive a `Wallet` as a parameter, then we can remove this check yep.

src/utils/ethereum.ts
```diff
@@ -34,32 +48,36 @@ async function getTokens(wallet, hermezTokens) {
...
-            // We can ignore if a call to the contract of a specific token fails.
-            .catch(() => {})
-        );
+        // We can ignore if a call to the contract of a specific token fails.
+        // ToDo: Find a way to properly type the functions of the contract declared through the ABI.
+        // eslint-disable-next-line
+        const balance: Promise<BigNumber> = contract
+          .balanceOf(ethereumAddress)
+          // eslint-disable-next-line @typescript-eslint/no-empty-function
+          .catch(() => {});
```
elias-garcia, line 65: To prevent disabling the `eslint` I think that you can do: `.catch(() => ({}));`

src/apis/price-updater.ts
```diff
@@ -24,14 +47,37 @@ const mockedFiatExchangeRates = {
   GBP: 0.716945,
 };
 
+const parsedReactAppPriceUpdaterApiUrl = z
+  .string()
+  .safeParse(process.env.REACT_APP_PRICE_UPDATER_API_URL);
```
elias-garcia, line 52: I think that we need to find a better way to parse the entire `env` object, if not we will need to do this every time we want to access them.

src/apis/price-updater.ts
```diff
@@ -1,20 +1,43 @@
...
+  price: z.number(),
+});
+
+const apiExchangeRateListParser: z.ZodSchema<ApiExchangeRate[]> = z.array(apiExchangeRateParser);
+
+interface ApiExchangeRateResponse {
+  currencies: ApiExchangeRate[];
+}
+
+const getFiatExchangeRatesResponseParser: z.ZodSchema<ApiExchangeRateResponse> = z.object({
```
elias-garcia, line 39: Same as before, shouldn't we move all this parsers and interfaces to persistence and the domain modules?

src/apis/hermez-web.ts
```diff
@@ -1,17 +1,31 @@
...
 function getNetworkStatus(): Promise<number> {
-  return axios
-    .get(`${baseApiUrl}/network-status.json`)
-    .then((res) => res.data.isWalletUnderMaintenance);
+  return axios.get(`${baseApiUrl}/network-status.json`).then((res) => {
+    const parsedResponse = walletUnderMaintenanceResponseParser.safeParse(res.data);
+    if (parsedResponse.success) {
+      return parsedResponse.data.isWalletUnderMaintenance;
+    } else {
+      return Promise.reject(parsedResponse.error.message);
```
elias-garcia, line 26: My idea here instead of `rejecting` is to return a custom `Promise` to avoid catching errors from the outside. Take a look to this function I wrote in Privado in the last days and let me know what you think: https://github.com/privadonetwork/privado-wallet/blob/main/src/persistence/api.ts#L121 Would be great to discuss different possibilities here.

.env.example
```diff
@@ -16,4 +16,6 @@ REACT_APP_WITHDRAWAL_DELAYER_CONTRACT_ADDRESS=0x8EEaea23686c319133a7cC110b840d15
 REACT_APP_BATCH_EXPLORER_URL=http://localhost:8080
 REACT_APP_ETHERSCAN_URL=https://etherscan.io
 REACT_APP_PRICE_UPDATER_API_URL=http://localhost:8080
-REACT_APP_PRICE_UPDATER_API_KEY=XXXXXXX
\ No newline at end of file
+REACT_APP_PRICE_UPDATER_API_KEY=XXXXXXX
+
+DISABLE_ESLINT_PLUGIN=true
```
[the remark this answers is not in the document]
amonsosanz: Definitely. Done.

src/apis/hermez-web.ts
```diff
@@ -1,17 +1,31 @@
 import axios from "axios";
+import * as z from "zod";
```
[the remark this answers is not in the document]
amonsosanz: I didn't know it was already exported as `z`! Changed!

src/apis/hermez-web.ts
amonsosanz: As [we agreed here](https://github.com/hermeznetwork/wallet-ui/pull/635#discussion_r701202023) everything under `/apis` needs to be moved to the `/persistence` namespace. I would though first migrate to TS and then reorganize the modules. Does it make sense?

> and WalletUnderMaintenanceResponse to the domain

I don't think the interfaces that represent responses from APIs belong to the `domain`. I think the domain layer should only include first-class interfaces that apply downward throughout the whole application.

src/utils/ethereum.ts
amonsosanz: Nice! Done

src/utils/ethereum.ts
amonsosanz: Nice trick! All they are now removed. Thx!

src/apis/price-updater.ts
amonsosanz: Worth considering, but for another slice. Some drawbacks of this approach may be:

 * A single decoding failure will make all the config decoding fail.
 * We will need to decode the config before the app initializes and pass the whole object as prop or context.

I don't see the current approach as a big problem, since all the parses will be under persistence. But let's give it a thought!

src/apis/price-updater.ts
amonsosanz: The same answer applies too.

src/apis/hermez-web.ts
amonsosanz: So you're basically wrapping on a box an operation that can either go well or fail, right?

src/apis/hermez-web.ts
elias-garcia: Right! I forgot about it. Both things makes sense to me

src/apis/price-updater.ts
elias-garcia: Yep, I said it just to keep it in mind for the next slices! Not to be covered now as we need to think if there are a better approach or not as you said!

src/apis/hermez-web.ts
elias-garcia: Hahahahahaha. Something like a mo... Yeah!

src/hermezjs-typings.d.ts
AlbertoElias: So Withdraw is the L1 part of the withdrawal process, so it is general enough, and hermezjs has a utility that takes care of the Withdraw.

And we are okay not supporting the other Transfer types, at least for now. If the use case comes up, we will add it. To prioritize features, we only add them when someone (or us) requests them

amonsosanz: Ok! Comment removed. Thx for the clarification.

.eslintrc.json
```diff
@@ -12,5 +12,8 @@
     "plugin:react/recommended",
     "plugin:react-hooks/recommended"
   ],
-  "ignorePatterns": ["/node_modules", "/build/", "/public/"]
+  "ignorePatterns": ["/node_modules", "/build/", "/public/", "**/*.js", "**/*.jsx"],
+  "rules": {
+    "@typescript-eslint/no-extra-semi": "off"
```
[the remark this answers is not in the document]
amonsosanz: I know... but this is how it behaves...

Well now we only disable it on one module, so I guess it's ok as it's, at least for now.

src/utils/ethereum.ts
amonsosanz: We've confirmed that `tx?.blockNumber` can be `null`. Comment updated.

AlbertoElias (approved):

src/views/account-details/account-details.view.jsx
amonsosanz: I made this decision for 2 reasons:

 * In this case, the view already has the props `fiatExchangeRates` and `preferredCurrency` so no new props need to be added to the view. 
 * If we access `fiatExchangeRates` from redux in the thunk, we need to test again the state of the async task before accessing its data.

Also, and this is subjective, I prefer functions to operate on their input parameters rather than accessing the state. It feels more pure and deterministic.

Anyway, if this does not fit with the style of the application I can change it. Should I?

src/apis/hermez-web.ts
amonsosanz: Don't say it!

src/hermezjs-typings.d.ts
amonsosanz: > Should we consider disabling this globally as we use this a lot including in catches?

No longer needed in catches! @elias-garcia showed me a nice trick to troll the linter:

`.catch(() => ({}))`

src/views/account-details/account-details.view.jsx
elias-garcia: Nope, it's perfectly fine. Just wanted to know the reason! I totally agree with this also: `I prefer functions to operate on their input parameters rather than accessing the state. It feels more pure and deterministic.`

PR #659, opened 2021-09-16
show invalid tag

### LINKED ISSUE #294: Give information on why tx is pending
se puede decir que fallo en algun momento, pero podria ser que se acabaraq forjando
motivos que podrian producir esta situacion:
- no se ha forjado (aun) una o mas txs con noce menor
- en ese momento por la conversion a USD la fee es demasiada pequeña
- en ese momento la account no tiene funds suficientes, pero puede recibir una tx que canvie esta situacion

there is an `info` property in /transactions-pool that gives the reason. We should read that and show it to the user in tx details

adrifdez:
Closes #294 

### What does this PR does?

If user receives a invalid transfer in the pool, we've change the tag to Invalid and inside the details as well
image

## Checklist

review them:

### DIFF
--- a/src/utils/accounts.js
+++ b/src/utils/accounts.js
@@ -22,7 +22,8 @@ function getAccountBalance(account, poolTransactions, pendingDeposits) {
 
   if (poolTransactions && poolTransactions.length) {
     const accountPoolTransactions = poolTransactions.filter(
-      (transaction) => transaction.fromAccountIndex === account.accountIndex
+      (transaction) =>
+        transaction.fromAccountIndex === account.accountIndex && !transaction.errorCode
     );
 
     accountPoolTransactions.forEach((pendingTransaction) => {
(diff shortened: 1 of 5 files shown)


AlbertoElias:
I think this is good, but some things to keep in mind:

- Did you validate this with @chamian?
- What about showing why a Tx is invalid?
- Currently, to show a more precise Account Balance, we add up pending txs from the pool. If a tx is invalid, it shouldn't count to that sum.

adrifdez:
> I think this is good, but some things to keep in mind:
> 
> * Did you validate this with @chamian?
> * What about showing why a Tx is invalid?
> * Currently, to show a more precise Account Balance, we add up pending txs from the pool. If a tx is invalid, it shouldn't count to that sum.

This was validated by Chami and she is going to validate the text coming in info field to included in the future in the details.
I'm going to change the sum to not add it to total balance when is invalid

amonsosanz (approved):

AlbertoElias (approved):

PR #661, opened 2021-09-22
Refactor primary button

### LINKED ISSUE #212: Refactor buttons
We have different kind of buttons across the app. We need to refactor them: naming, styles, make them more reusable...

adrifdez:
Closes #212 

### What does this PR does?

Refactor FormButton to PrimaryButton to reuse styles inside of boxes.

## Checklist

review them:


src/views/shared/button/button.view.jsx
```diff
@@ -6,22 +6,19 @@ import useButtonStyles from "./button.styles";
 
 function Button({ Icon, text, className, disabled, onClick }) {
   const classes = useButtonStyles({ rounded: !text });
-  const isClickable = onClick !== undefined;
```
AlbertoElias, line 9: Why stop supporting this behaviour?

AlbertoElias: And where is this button used?

AlbertoElias (commented):
Couple questions, rest looks good!

src/views/shared/button/button.view.jsx
adrifdez: Because I think that a Button component must have an onClick prop and should be clickable.
image
image

AlbertoElias: If all instances are clickable and it doesn't break anything, then great!

AlbertoElias (approved):

src/views/shared/button/button.view.jsx
adrifdez: Yes, all components are passing that prop

PR #663, opened 2021-09-22
Refactor to typescript (slice 04: store/login)

### LINKED ISSUE #662: Refactor to Typescript (slice 04)

amonsosanz:
Closes #662.

### What does this PR does?

This 4th slice is mainly focused on the store/login modules.

## Checklist

review them:


src/store/global/global.thunks.ts
```diff
@@ -10,21 +10,17 @@ import { push } from "connected-react-router";
 import { ethers } from "ethers";
 import { Block, TransactionReceipt, TransactionResponse } from "@ethersproject/providers";
 import Connector from "@walletconnect/web3-provider";
-
 import HermezABI from "@hermeznetwork/hermezjs/src/abis/HermezABI";
 import { TxType, TxState } from "@hermeznetwork/hermezjs/src/enums";
-
```
AlbertoElias, line 16: We do keep the separation between external imports and internal ones

src/store/login/login.reducer.ts
```diff
@@ -0,0 +1,245 @@
...
+import { AsyncTask } from "src/utils/types";
+// domain
+import { HermezWallet } from "src/domain/hermez";
+import { AuthSignatures } from "src/domain/local-storage";
+import { getAuthSignatures } from "src/persistence/local-storage";
+
+export interface LoginState {
+  step: Step;
+  // ToDo: This state is never set. The actions do not exist.
+  // networkNameTask: AsyncTask<string, string>;
```
AlbertoElias, line 16: If we don't use it anymore, maybe it can be removed? Always happy to remove things hehe

src/store/login/login.reducer.ts
```diff
@@ -0,0 +1,245 @@
...
+  WALLET_SELECTOR = "wallet-selector",
+  ACCOUNT_SELECTOR = "account-selector",
+  WALLET_LOADER = "wallet-loader",
+  CREATE_ACCOUNT_AUTH = "create-account-auth",
+}
+
+export type Step =
+  | {
+      type: "wallet-selector";
+      // ToDo: Why don't we move walletName to LoginState? It's in three steps and can be undefined...
```
AlbertoElias, line 32: Sounds good to me

src/store/login/login.reducer.ts
```diff
@@ -0,0 +1,245 @@
...
+      walletName: WalletName | undefined;
+    }
+  | {
+      type: "account-selector";
+      // ToDo: What is this for? Isn't it the same as wallet-selector?
```
AlbertoElias, line 37: This is remains from hardware wallet stuff

src/store/login/login.thunks.ts
```diff
@@ -0,0 +1,305 @@
...
+    }
+  }
+}
+
+/**
+ * Helper function that signs the authentication message depending on Wallet type
+ * @param {Web3Provider | Signer} providerOrSigner
+ * @param {String} message
+ * @param {String} address
+ * @returns {Promise<unknown>} A promise that resolves to the signature
```
AlbertoElias, line 70: Isn't this Promise<string>? Also, with TS, do we need params and returns docs?

src/store/login/login.reducer.ts
amonsosanz: Yeah me too, I was afraid to remove it...

src/store/global/global.thunks.ts
amonsosanz: Wait, then does this imply that we keep a certain order for imports? This is getting complicated...

src/store/login/login.thunks.ts
amonsosanz: This should be `Promise<string>`.

> Also, with TS, do we need params and returns docs?

I'm glad you mention it! No!! JSDocs are completely useless now and even counterproductive since they use to get obsoleted and lie to us.

Can I remove them as I go?

src/store/login/login.reducer.ts
amonsosanz: Removed

src/store/login/login.reducer.ts
amonsosanz: I won't do it now though. It's better to do it after moving `views/login/login.view.jsx` to TS.

src/store/login/login.reducer.ts
amonsosanz: Ok, we can remove it when we get rid of it then.

src/store/global/global.thunks.ts
AlbertoElias: Basically it's

- External Imports (from npm dependencies)
- New Line
- Internal Imports

src/store/login/login.thunks.ts
AlbertoElias: I would remove only params and returns but keep function descriptions. But yeah, go ahead :)

AlbertoElias (approved):

adrifdez (approved):

src/store/login/login.thunks.ts
amonsosanz: Cool!

src/store/global/global.thunks.ts
amonsosanz: Ok. Done!

adrifdez (approved):

AlbertoElias (approved):

src/store/login/login.reducer.ts
elias-garcia: Mmmm not sure why it can be `undefined`, because the `walletName` should exist only after clicking uour login method (metamask, or wallet connect). Maybe it doesn't make sense to have it in this step only, but maybe in the other ones, but without the possibility to be `undefined`.

amonsosanz: I've removed `walletName` from `wallet-selector` and `account-selector` login steps, and removed the `undefined` from the `wallet-loader` step.

src/store/login/login.thunks.ts
```diff
@@ -0,0 +1,306 @@
...
+      const {
+        login: { step },
+      } = getState();
+
+      if (step.type === "wallet-loader") {
+        dispatch(globalActions.loadWallet(wallet));
+        // ToDo: SignerData does not have the prop "address" on all its shapes. Why do we need to add the address?
+        //       For now I'm creating an IIFE to set the address where it makes sense, but we should confirm whether
+        //       this is necessary or not. Am I messing it up with the wrong types?
+        const signerDataWithAddress: SignerData = ((): SignerData => {
```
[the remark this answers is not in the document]
amonsosanz: > Mmmm, I think that the `address` is not needed in the store, we only need to hold the `SignerData`

Agreed. We only save `SignerData` on the global store.

> Here you can find the data needed in `hermezjs` for each type, if you take a look only the `TREZOR` case needs the address.

Yep, from there I got the info to define the type [`SignerData`](https://github.com/hermeznetwork/wallet-ui/blob/refactor-to-typescript-slice-04/src/hermezjs-typings.d.ts#L665). Notice that the member `JsonRpcSignerData` of the union can also hold the address on its `addressOrIndex` prop. Both props are optional though.

> so maybe it's better to build properly the `signerData` in the `getSignerData` function

I'm not sure I'm getting what you propose here. To properly build the `SignerData` (i.e include the address) in the `getSignerData` function, we need to pass the address to the function. To pass the address to the function we would need to call `getSignerData` twice, one initial time without address and then again with the address. Is that what you mean?

I may be lost, but otherwise, I think that the current implementation (IIFE) is better.

> to later store the address in Redux

Lost here.

Maybe we can meet about this...

PR #668, opened 2021-09-24
Back to selectToken

### LINKED ISSUE #471: Add functionality of changing selected token in Send/Deposit/Withdraw form
Allow the user to go back to the list of available tokens when clicking token component.

change selected token

adrifdez:
Closes #471 

### What does this PR does?

Allow users to go to token selector when they are building a transaction

## Checklist

review them:


amonsosanz:
After a while of investigating I still haven't found the root of this error:

error-changing-token-on-tx

It's probably related to setting the state to an unmounted component, but we need to find what component is it so we can protect the case.

The error is only shown the first time you change the token after reloading the page.

amonsosanz (dismissed):

src/views/shared/amount-input/amount-input.view.jsx
```diff
@@ -25,9 +25,19 @@ function AmountInput(Component) {
...
         .getGasPrice()
-        .then((gasPrice) => setGasPrice(gasPrice));
+        .then((gasPrice) => {
+          if (isActive) {
+            setGasPrice(gasPrice);
+          }
+        });
+
+      return () => {
+        isActive = false;
```
AlbertoElias, line 39: I don't get this, what is it for?

amonsosanz: Indeed, it's ugly but has been added to prevent [this error](https://github.com/hermeznetwork/wallet-ui/pull/668#issuecomment-929298899) which was occurring because we were attempting to set a state on an unmounted component. We should definitely find a better approach than this mutable flag. More info [here](https://0xpolygon.slack.com/archives/C02FS3PHEGH/p1632840307005600).

AlbertoElias: I'm no React expert. What other options do we have?

adrifdez: I'm refactoring for other pr and I'm going to remove this effect

AlbertoElias: Do you mean you are going to remove this in a different PR? Which one?

adrifdez: My last change in https://github.com/hermeznetwork/wallet-ui/pull/665 removes that effect

AlbertoElias: So can it be removed here then?

adrifdez: it's going to create a conflict so I'm waiting to close the other pr to solve the conflict here

AlbertoElias (approved):

PR #683, opened 2021-09-29
Refactor to typescript slice 05

### LINKED ISSUE #682: Refactor to Typescript (slice 05)

amonsosanz:
Closes #682.

### What does this PR does?

This slice mainly focuses on migrating to Typescript the namespace `store/my-account`.

## Checklist

review them:


src/persistence/local-storage/index.ts
```diff
@@ -298,14 +315,14 @@ export function removePendingDepositByHash(
       [hermezEthereumAddress]: deposits.filter((deposit) => deposit.hash !== hash),
     },
   };
-  setStorageKey(constants.PENDING_DEPOSITS_KEY, newStorage);
+  setStorageKey(constants.PENDING_DEPOSITS_KEY, JSON.stringify(newStorage));
```
AlbertoElias, line 318: Might bee worth reconsidering this. It kinda sucks we can't abstract it just for 1 storage element that isn't an object

amonsosanz: I think it's ok as it's now, `setStorageKey` accepts a `string` so if you already have a string you can send it straight, if you have an object you use `JSON.stringify` and if you have a number you can use `toString`.

> It kinda sucks we can't abstract it just for 1 storage element that isn't an object

We don't need to abstract it, having just 1 storage element that isn't an object may change tomorrow. This feels like the correct implementation to me.

AlbertoElias: As you see better :)

AlbertoElias (approved):

AlbertoElias (approved):

src/persistence/local-storage/index.ts
```diff
@@ -30,14 +30,14 @@ export function getStorage(key: string): unknown {
   }
 }
 
-export function setStorageKey(key: string, storage: unknown): void {
-  localStorage.setItem(key, JSON.stringify(storage));
+export function setStorageKey(key: string, value: string): void {
+  localStorage.setItem(key, value);
```
elias-garcia, line 34: Why not just using a union type and keep the `JSON.stringify` centralized in one place?

```
export function setStorageKey(key: string, value: Record<string, unknown> | string): void {
  if (typeof value === "string") {
    localStorage.setItem(key, value);
  } else {
    localStorage.setItem(key, JSON.stringify(value));
  }
}
```

src/persistence/local-storage/index.ts
```diff
@@ -17,7 +17,7 @@ import {
 
 // Storage Helpers
 
-export function getStorage(key: string): unknown {
+export function getStorageKey(key: string): unknown {
```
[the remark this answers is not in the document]
amonsosanz: Ok. Done.

src/persistence/local-storage/index.ts
amonsosanz: I'm not sure about this, mainly because `Record<string, unknown> | string` is only a subset of the shapes that `value` could adopt. What if we want to store a `number` or a `boolean`? Then, we still would need to convert them to string in the origin. In that case, why is a Record special?

We could make this union wider, and manage each case, but we would eventually be covering `unknown`.

I prefer the current implementation, but if you 2 still prefer centralizing the `JSON.stringify` here, let me know and I do the change.

amonsosanz: > JSON.stringify parses booleans and numbers just fine

Yeah, this is indeed true. Ok, agreed and done!

adrifdez (approved):

PR #693, opened 2021-10-05
Refactor to typescript slice 06

### LINKED ISSUE #684: Refactor to Typescript (slice 06)

amonsosanz:
Closes #684

### What does this PR does?

This slice will start the migration to Typescript of the namespaces `store/transaction` and `views/transaction`.

After analyzing the code on these namespaces it has been concluded that it is important to decouple the different types of transactions supported by the UI. i.e Send, Deposit, and Withdraw, so we end up with a set of [action/reducer/thunks/view] per each type of transaction.

The main reason for this restructuration is that the current code is tangled, complex, and difficult to maintain. 

It's important though that the components are kept generic, so they can be used by the view of any transaction type.

At the store level, this slice will refactor the transactions of type `Send`, creating the following modules in the store (notice the plural in `transactions`):

  * `src/store/transactions/transfer/transfer.actions.ts`
  * `src/store/transactions/transfer/transfer.reducer.ts`
  * `src/store/transactions/transfer/transfer.thunks.ts`

At the presentation level, this slice will only cover the view `transactions/transfer` in charge of managing the transactions of type `Send`. Apart from the view itself, this slice must also include the refactor of the components used by the view.

ℹ️ It's important to note that the old namespaces `store/transaction` and `views/transaction` (singular transaction) **won't be modified at all** in this refactor. This ensures that the types of transactions not covered by each slice will keep working as always.

Once all the new `store` and `views` namespaces are migrated we will be able to completely remove the old namespaces `store/transaction` and `views/transaction`.

### How to test?

In this slice, testing should focus especially on the transactions of type "Send" (a.k.a "Transfer") which are deeply refactored.

## Checklist

review them:

- [*] Respect code style and lint


src/views/transactions/components/transaction-overview/transaction-overview.view.tsx
```diff
@@ -0,0 +1,337 @@
+// ToDo: Remove the disable of TS and the linter below once the component are migrated to TS
```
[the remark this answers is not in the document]
AlbertoElias: Oh cool, thanks for the info! Might be good to get this out of the way after Tx Details

src/hermezjs-typings.d.ts
```diff
@@ -90,14 +91,17 @@ declare module "@hermeznetwork/*" {
     // ToDo: This states are yet to be validated
     // state: "fged" | "fing" | "pend" | "invl";
     // ToDo: Are there more deposit types??
-    type: "Deposit" | "CreateAccountDeposit";
+    type: typeof TxType.Deposit | typeof TxType.CreateAccountDeposit;
```
elias-garcia, line 94: `typeof` is not needed here

src/hermezjs-typings.d.ts
```diff
@@ -11,6 +11,7 @@
 
 declare module "@hermeznetwork/*" {
   import { BigNumber } from "ethers";
+  import { TxTypesUnion, TxStatesUnion } from "@hermeznetwork/hermezjs/src/enums";
```
elias-garcia, line 14: Why the `Union` suffix? I think that it would make more sense to call them just `TxType` and `TxState`.

src/views/transactions/components/transaction-overview/transaction-overview.view.tsx
```diff
@@ -0,0 +1,337 @@
...
+import { HermezWallet, Account, FiatExchangeRates, Exit } from "src/domain/hermez";
+
+interface EstimatedWithdrawFee {
+  amount: string;
+  USD: number;
+}
+
+type Transaction =
+  | {
+      type: typeof TxType.Deposit;
```
elias-garcia, line 33: `typeof` is not needed here neither

src/views/transactions/components/transaction-overview/transaction-overview.view.tsx
```diff
@@ -0,0 +1,337 @@
...
+      ) => void;
+    }
+  | {
+      type: typeof TxType.ForceExit;
+      amount: BigNumber;
+      account: Account;
+      onForceExit: (amount: BigNumber, account: Account) => void;
+    };
+
+interface Props {
```
elias-garcia, line 76: Maybe `TransactionOverviewProps`?

src/views/transactions/components/transaction-overview/transaction-overview.view.tsx
```diff
@@ -0,0 +1,337 @@
...
+      type: typeof TxType.Deposit;
+      amount: BigNumber;
+      account: Account;
+      onDeposit: (amount: BigNumber, account: Account) => void;
+    }
+  | {
+      type: typeof TxType.Transfer;
+      amount: BigNumber;
+      account: Account;
+      to: Partial<Account>;
```
elias-garcia, line 42: Can we avoid to have this `Partial<Account>` and have another `Account` here to be more domain driven?

src/hermezjs-typings.d.ts
amonsosanz: It's required because `TxType.Deposit` and `TxType.CreateAccountDeposit` are string literals, not types.  Without `typeof` TS infers the types as `any`:

typeof

src/persistence/parsers/index.ts
```diff
@@ -35,6 +35,8 @@ const account: z.ZodSchema<Account> = z.object({
   accountIndex: z.string(),
   balance: z.string(),
   bjj: z.string(),
+  hezEthereumAddress: z.string(),
+  hezBjjAddress: z.string().optional(),
```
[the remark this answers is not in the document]
amonsosanz: I know. I added a ToDo comment to discuss this:

image

Should I then add it to the Domain?

src/hermezjs-typings.d.ts
```diff
@@ -703,19 +697,35 @@ declare module "@hermeznetwork/hermezjs/src/enums" {
     Withdraw = "Withdrawn",
     Exit = "Exit",
     ForceExit = "ForceExit",
-  };
-
-  declare const TxState = {
+  } as const;
+
+  export type TxTypesUnion =
```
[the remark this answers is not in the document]
amonsosanz: Precisely because `enums` are both types and values, they can not be used in TypeScript Declaration files (`.d.ts`).

I've learnt this in this PR when I realized I was having problems in runtime when importing the enums to use them as values. They are never available as values at runtime. So I switched to [const assertions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions) so we can keep using these definitions as both values and types (using `typeof`).

I also decided to export the unions `TxTypesUnion` and `TxStatesUnion` because they are used in several places.

Some more info on the topic: https://lukasbehal.com/2017-05-22-enums-in-declaration-files/

src/views/transactions/components/transaction-confirmation/transaction-confirmation.view.jsx
```diff
@@ -0,0 +1,51 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { TxType } from "@hermeznetwork/hermezjs/src/enums";
+
+import useTransactionConfirmationStyles from "./transaction-confirmation.styles";
+import transactionConfirmation from "../../../../images/transaction-confirmation.svg";
+import PrimaryButton from "../../../shared/primary-button/primary-button.view";
+
+function TransactionConfirmation({ transactionType, onFinishTransaction }) {
```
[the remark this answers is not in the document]
amonsosanz: `TransactionConfirmation` was still imported but not used by the `transaction.view.jsx`. I've removed the import and the component.

src/views/transactions/components/transaction-error/transaction-error.view.jsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+import useTransactionErrorStyles from "./transaction-error.styles";
+import transactionError from "../../../../images/transaction-error.png";
+import PrimaryButton from "../../../shared/primary-button/primary-button.view";
+
+function TransactionError({ onFinishTransaction }) {
```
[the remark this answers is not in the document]
amonsosanz: `TransactionError` was still imported but not used by the `transaction.view.jsx`. I've removed the import and the component.

src/views/transactions/components/transaction-overview/transaction-overview.view.tsx
amonsosanz: The [same response](#discussion_r726032975) applies.

src/views/transactions/components/transaction-overview/transaction-overview.view.tsx
amonsosanz: Of course. Done!

src/views/transactions/components/transaction-overview/transaction-overview.view.tsx
amonsosanz: mmm... I'm not sure I follow. What do you mean by having another `Account`? Why would we need to declare another interface for an account? I'm probably missing something. Can you elaborate a bit more?

src/store/transactions/transfer/transfer.thunks.ts
```diff
@@ -0,0 +1,184 @@
...
+
+    if (
+      wallet !== undefined &&
+      (coordinatorStateTask.status === "successful" || coordinatorStateTask.status === "reloading")
+    ) {
+      dispatch(transferActions.startTransactionApproval());
+
+      const nextForgerUrls = getNextForgerUrls(coordinatorStateTask.data);
+      const txData = {
+        type: "Transfer" as const,
```
elias-garcia, line 155: This `as const` is kinda weird also, this is because of the enum setup in the declarations file right?

src/views/transactions/components/account-selector/account-selector.view.jsx
```diff
@@ -18,14 +18,19 @@ function AccountSelector({
...
     .filter((deposit) => deposit.type === TxType.CreateAccountDeposit)
     .map((deposit) => deposit.token.id);
 
   React.useEffect(() => {
     if (accountsTask.status === "pending" && pooledTransactionsTask.status === "successful") {
-      onLoadAccounts(undefined, pooledTransactionsTask.data, fiatExchangeRates, preferredCurrency);
+      onLoadAccounts(
+        undefined,
+        pooledTransactionsTask.data,
+        pendingDeposits,
```
adrifdez, line 30: `onLoadAccounts` doesn't need `pendingDeposit`. Now it's not expecting that value

src/persistence/parsers/index.ts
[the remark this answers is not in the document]
elias-garcia: Yep, I would remove this property from the account domain

src/hermezjs-typings.d.ts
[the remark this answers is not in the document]
amonsosanz: The problems I had were:

  * Using string literals as values for enum types
  * Naming the enums differently than the consts exported by the lib (TxType and TxState)

Fixed!

src/hermezjs-typings.d.ts
amonsosanz: Now with enums, the `typeof` has been removed.

PR #695, opened 2021-10-06
Improve how we check own wallet

### LINKED ISSUE #688: Clicking on a pending deposit in Account details redirects to /
It tries to navigate to `/accounts/${accountIndex}/transactions/${transactionId}` but ends up going to `/`

adrifdez:
Closes #688 

### What does this PR does?

We've improved how we check if the transaction details owns to user wallet

## Checklist

review them:

### DIFF
--- a/src/store/transaction-details/transaction-details.thunks.js
+++ b/src/store/transaction-details/transaction-details.thunks.js
@@ -47,7 +47,12 @@ function fetchTransaction(transactionIdOrHash) {
 
     return transactionPromise
       .then((res) => {
-        if (res.fromBJJ !== wallet.publicKeyBase64 && res.toBJJ !== wallet.publicKeyBase64) {
+        if (
+          res.fromBJJ !== wallet.publicKeyBase64 &&
+          res.toBJJ !== wallet.publicKeyBase64 &&
+          res.fromHezEthereumAddress !== wallet.hermezEthereumAddress &&
+          res.toHezEthereumAddress !== wallet.hermezEthereumAddress
+        ) {
           dispatch(push("/"));
         } else {
           dispatch(transactionDetailsActionTypes.loadTransactionSuccess(res));
--- a/src/store/transaction-details/transaction-details.thunks.js
+++ b/src/store/transaction-details/transaction-details.thunks.js
@@ -48,15 +48,15 @@ function fetchTransaction(transactionIdOrHash) {
     return transactionPromise
       .then((res) => {
         const txContainsBJJUserAddress =
-          res.fromBJJ !== wallet.publicKeyBase64 && res.toBJJ !== wallet.publicKeyBase64;
+          res.fromBJJ === wallet.publicKeyBase64 || res.toBJJ === wallet.publicKeyBase64;
         const txContainsHezEthereumAddressUserValue =
-          res.fromHezEthereumAddress !== wallet.hermezEthereumAddress &&
-          res.toHezEthereumAddress !== wallet.hermezEthereumAddress;
+          res.fromHezEthereumAddress === wallet.hermezEthereumAddress ||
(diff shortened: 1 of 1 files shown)


src/store/transaction-details/transaction-details.thunks.js
```diff
@@ -47,7 +47,12 @@ function fetchTransaction(transactionIdOrHash) {
 
     return transactionPromise
       .then((res) => {
-        if (res.fromBJJ !== wallet.publicKeyBase64 && res.toBJJ !== wallet.publicKeyBase64) {
+        if (
+          res.fromBJJ !== wallet.publicKeyBase64 &&
+          res.toBJJ !== wallet.publicKeyBase64 &&
+          res.fromHezEthereumAddress !== wallet.hermezEthereumAddress &&
+          res.toHezEthereumAddress !== wallet.hermezEthereumAddress
+        ) {
```
amonsosanz, line 55: I think we should take this opportunity to improve the semantics of this conditional. The conditions above are hard to follow for newcomers.

Imagine we abstract these 4 conditions to variables:

```typescript
const A = res.fromBJJ !== wallet.publicKeyBase64;
const B = res.toBJJ !== wallet.publicKeyBase64;
const C = res.fromHezEthereumAddress !== wallet.hermezEthereumAddress;
const D = res.toHezEthereumAddress !== wallet.hermezEthereumAddress;
```

How would you name the variables `A`, `B`, `C`, and `D`?

src/store/transaction-details/transaction-details.thunks.js
```diff
@@ -47,7 +47,13 @@ function fetchTransaction(transactionIdOrHash) {
...
     return transactionPromise
       .then((res) => {
-        if (res.fromBJJ !== wallet.publicKeyBase64 && res.toBJJ !== wallet.publicKeyBase64) {
+        const txContainsBJJUserAddress =
+          res.fromBJJ !== wallet.publicKeyBase64 && res.toBJJ !== wallet.publicKeyBase64;
+        const txContainsHezEthereumAddressUserValue =
+          res.fromHezEthereumAddress !== wallet.hermezEthereumAddress &&
+          res.toHezEthereumAddress !== wallet.hermezEthereumAddress;
+
+        if (isHezEthereumAddressUserWallet && isHezEthereumAddressUserWallet) {
```
amonsosanz, line 56: I love the names of the vars! It would be great though using them in the conditional! 😅 What is `isHezEthereumAddressUserWallet`?

Something else: Isn't now the logic of each var inverted? I mean, `txContainsBJJUserAddress` means that the transaction DOES contain the user address, so it feels the condition should be `res.fromBJJ === wallet.publicKeyBase64 || res.toBJJ === wallet.publicKeyBase64`.

On the other hand, the right name for the condition `res.fromBJJ !== wallet.publicKeyBase64 && res.toBJJ !== wallet.publicKeyBase64` is more like `txDoesNotContainsBJJUserAddress`.

Does this make sense?

Since truthy conditions help with human comprehension, I prefer the option:
```typescript
const txContainsBJJUserAddress = res.fromBJJ === wallet.publicKeyBase64 || res.toBJJ === wallet.publicKeyBase64;
```
Then, assuming we only had to check this condition, the `if` could be:
```typescript
if (txContainsBJJUserAddress === false) { // we send the user to "/" }
```

But we need to check 2 conditions and regarding this I think: shouldn't we send the user to `"/"` **either** when `txContainsBJJUserAddress` or `txContainsHezEthereumAddressUserValue` is `true`?

amonsosanz: It's perfect now! Much easier to follow and reason about!

amonsosanz (approved):

AlbertoElias (approved):

PR #698, opened 2021-10-11
add untouched value to AmountInput

### LINKED ISSUE #692: Avoid send TX without amount
When I initiate a Send (transfer) transaction, the **Continue** button is initially disabled:

image

I can then paste an address and the **Continue** button gets activated even when an amount has not yet been provided:

image

Allowing to sending an amount of 0, which obviously fails:

image

adrifdez:
Closes #692 

### How to test?

User needs to insert a valid amount to see the button available to send a transfer to an account

## Checklist

review them:

### DIFF
--- a/src/views/transaction/components/transaction-form/transaction-form.view.jsx
+++ b/src/views/transaction/components/transaction-form/transaction-form.view.jsx
@@ -60,6 +60,7 @@ function TransactionForm({
     React.useState(undefined);
   const [gasPrice, setGasPrice] = React.useState(BigNumber.from(0));
   const [depositFee, setDepositFee] = React.useState(undefined);
+  const [isAmountInputDirty, setIsAmountInputDirty] = React.useState(false);
 
   React.useEffect(() => {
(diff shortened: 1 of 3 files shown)


src/views/transaction/components/transaction-form/transaction-form.view.jsx
```diff
@@ -165,8 +165,11 @@ function TransactionForm({
   }
 
   function handleAmountChange(data) {
-    setAmount(data.amount.tokens);
     setShowInFiat(data.showInFiat);
+
+    if (data.isUntouched) return;
```
elias-garcia, line 171: Instead of creating more branches in the code (if...else) which makes it more difficult to later find bugs, I will create a `const [isAmountInputDirty, setIsAmountInputDirty] = React.useState()` and then here just set the value: `setIsAmountInputDirty(data.isDirty)`. Later when you decide to display the error or not you just render it in case `isAmountInputDirty === true && isAmountInvalid === true`.

AlbertoElias (approved):

PR #713, opened 2021-10-21
move color to body

### LINKED ISSUE #709: Check the black colors in the UI

adrifdez:
Closes #709 
### What does this PR does?

Move color property from * to body to allow the inherited behaviour

## Checklist

review them:

### DIFF
--- a/src/views/app.styles.js
+++ b/src/views/app.styles.js
@@ -46,7 +46,6 @@ const useAppStyles = createUseStyles((theme) => ({
   "@global": {
     "*": {
       boxSizing: "border-box",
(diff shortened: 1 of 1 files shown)


src/views/app.styles.js
```diff
@@ -46,14 +46,14 @@ const useAppStyles = createUseStyles((theme) => ({
   "@global": {
     "*": {
       boxSizing: "border-box",
```
elias-garcia, line 48: Maybe is also a good idea to move the `border-box`?

adrifdez: box-sizing is not inherited

PR #723, opened 2021-10-28
Refactor to typescript slice 11 (deposit tx)

### LINKED ISSUE #720: Refactor to Typescript (slice 11)
This slice is focused on refactoring to Typescript the Deposit transaction.

amonsosanz:
Closes #720.

### What does this PR does?

This slice is focused on refactoring to Typescript the Deposit transaction.

## Checklist

review them:


src/views/transactions/components/account-selector/account-selector.view.jsx
```diff
@@ -25,7 +25,11 @@ function AccountSelector({
 
   React.useEffect(() => {
     if (accountsTask.status === "pending" && poolTransactionsTask.status === "successful") {
-      onLoadAccounts(undefined, poolTransactionsTask.data, fiatExchangeRates, preferredCurrency);
+      if (transactionType === TxType.Deposit) {
+        onLoadAccounts(fiatExchangeRates, preferredCurrency);
```
elias-garcia, line 29: If we are loading the accounts without the `poolTransactionsTask` why put it inside the parent `if`? I'll put it outside and then we can remove the `poolTransactionsTask` from the reducer and its actions I think.

src/views/transactions/components/transaction-form/transaction-form.view.jsx
```diff
@@ -260,7 +263,7 @@ function TransactionForm({
         const accountChecks = [
           getAccounts(receiver, [account.token.id]),
           ...(!isHermezBjjAddress(receiver)
-            ? [getCreateAccountAuthorization(receiver).catch(() => undefined)]
+            ? [getCreateAccountAuthorization(receiver).catch(() => ({}))]
```
elias-garcia, line 266: This is breaking the create authorization check in the form! If we return something different than `undefined` or `null` the check is going to break and will allow to send the tx without having the create account auth signature.

src/domain/index.ts
```diff
@@ -34,3 +34,9 @@ export interface EstimatedWithdrawFee {
   amount: BigNumber;
   USD: number;
 }
+
+export interface EstimatedDepositFee {
+  amount: string;
+  gasPrice: BigNumber | null;
```
elias-garcia, line 40: Taking a look at the code in the thunk we can remove this `null` from here as you are already checking it in the interface.

src/views/transactions/components/transaction-form/transaction-form.view.jsx
```diff
@@ -69,7 +70,9 @@ function TransactionForm({
 
   React.useEffect(() => {
     if (estimatedDepositFeeTask.status === "successful") {
-      setGasPrice(estimatedDepositFeeTask.data.gasPrice);
+      if (estimatedDepositFeeTask.data.gasPrice !== null) {
```
elias-garcia, line 73: After changing the interface we can also remove this check I think.

src/hermezjs-typings.d.ts
```diff
@@ -280,6 +278,10 @@ declare module "@hermeznetwork/*" {
     pendingItems: number;
   }
 
+  export interface Tokens {
+    tokens: Token[];
+    pendingItems: number;
+  }
```
elias-garcia, line 284: Missing new line below

elias-garcia (changes requested):

src/store/transactions/deposit/deposit.thunks.ts
```diff
@@ -0,0 +1,251 @@
...
+      );
+    }
+  };
+}
+
+/**
+ * Fetches the recommended fees from the Coordinator
+ * @returns {void}
+ */
+function fetchFees(): AppThunk {
```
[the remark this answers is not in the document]
adrifdez: Yes onLoadFees is just used on the view when is a Exit

src/views/transactions/components/transaction-form/transaction-form.view.jsx
adrifdez: I think that we need to review this logic deeper

elias-garcia: Yep, we will do it when we refactor this component

src/store/transactions/deposit/deposit.thunks.ts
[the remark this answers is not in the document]
amonsosanz: Interesting!

> Yes onLoadFees is just used on the view when is a Exit

This is not strictly true:

image

Can I then move `onLoadFees()` to the `if` so it's only executed when `transactionType === TxType.Exit`?

Otherwise, I'm forced to pass some dummy function, which is ugly.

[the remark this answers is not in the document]
elias-garcia: No! Fees are fetched for L2 txs only, so we need to call the `onLoadFees` for transactions and for exits.

src/views/transactions/components/account-selector/account-selector.view.jsx
amonsosanz: Done

src/views/transactions/components/transaction-form/transaction-form.view.jsx
amonsosanz: I [explicitly asked](https://0xpolygon.slack.com/archives/C02FS3PHEGH/p1635513098067600) about this and [you agreed](https://0xpolygon.slack.com/archives/C02FS3PHEGH/p1635518824074300) to leave the `{}`.

Then it needs to be `undefined` because we're checking for a falsy value. Ok!

amonsosanz: Done. Added a comment to clarify this.

src/domain/index.ts
amonsosanz: Right! Done.

src/views/transactions/components/transaction-form/transaction-form.view.jsx
amonsosanz: Right! Done.

src/hermezjs-typings.d.ts
amonsosanz: Added!

src/views/transactions/components/transaction-form/transaction-form.view.jsx
elias-garcia: Yep, I thought that you were talking about a different thing, my fault, sorry!

src/store/transactions/deposit/deposit.thunks.ts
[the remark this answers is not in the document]
amonsosanz: Done!

elias-garcia (approved):

PR #727, opened 2021-11-08
Get the proper type of transfer

### LINKED ISSUE #725: Transfer (at least with MATIC) not working in mainnet
When sending MATIC to a valid Hez address, in Tx Details the From and To appear as the same. And when you click send, it fails. Probably worth checking with both MATIC and other tokens and see if the refactor has fixed this. We might need to deploy MATIC in testnet.

adrifdez:
Closes #725 

### What does this PR does?

We get the proper type of the transfer to send properly the address in the specific field

## Checklist

review them:

### DIFF
--- a/src/hermezjs-typings.d.ts
+++ b/src/hermezjs-typings.d.ts
@@ -384,10 +384,7 @@ declare module "@hermeznetwork/hermezjs/src/tx" {
 declare module "@hermeznetwork/hermezjs/src/tx-utils" {
   import { ScalarValue } from "@hermeznetwork/hermezjs";
   import { TxType } from "@hermeznetwork/hermezjs/src/enums";
-
-  interface Transaction {
-    to: string | undefined;
-  }
+  import { Tx } from "@hermeznetwork/hermezjs/src/tx";
 
   // function _encodeTransaction();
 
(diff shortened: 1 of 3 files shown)


src/hermezjs-typings.d.ts
```diff
@@ -395,7 +400,8 @@ declare module "@hermeznetwork/hermezjs/src/tx-utils" {
 
   function getMaxAmountFromMinimumFee(minimumFee: ScalarValue, balance: ScalarValue): ScalarValue;
 
-  // function getTransactionType();
+  function getTransactionType(transaction: Transaction): TxType;
```
elias-garcia, line 403: We already have an interface for transactions which is `Tx`. I know that in this case we just need the `to` from the `Tx` interface, so I would suggest to remove this new `Transaction` interface and change this signature to:

```ts
function getTransactionType(transaction: { to: Pick<Tx, "to"> }): TxType;
```

This is going to pick the same type of the `to` field in the already existing `Tx` interface without duplicating it.

elias-garcia: Sorry @adrifdez, forgot about it's holiday in your city today. I'll change it and merge it.

elias-garcia (approved):

elias-garcia (approved):

amonsosanz (approved):

PR #734, opened 2021-11-10
Fix wrong message shown in the exit loader

### LINKED ISSUE #729: Fix wrong message shown in the exit loader
When you execute step 1 of a withdrawal (exit) or a transfer (send), the loader asks you to sign the TX in the wallet, although you don't actually have to sign it.

sign-tx-message-bug

amonsosanz:
Closes #729.

### What does this PR does?

When you execute step 1 of a withdrawal (exit) or a transfer (send), the loader asks you to sign the TX in the wallet, although you don't actually have to sign it.

## Checklist

review them:

### DIFF
--- a/src/views/transactions/components/transaction-overview/transaction-overview.view.tsx
+++ b/src/views/transactions/components/transaction-overview/transaction-overview.view.tsx
@@ -170,15 +170,18 @@ function TransactionOverview({
     </Container>
   );
 
-  const footer = (buttonLabel: string): JSX.Element =>
+  const footer = (buttonLabel: string, loaderText: string): JSX.Element =>
     isTransactionBeingApproved ? (
       <div className={classes.signingSpinnerWrapper}>
         <Spinner />
-        <p className={classes.signingText}>Sign in your connected wallet to confirm transaction</p>
+        <p className={classes.signingText}>{loaderText}</p>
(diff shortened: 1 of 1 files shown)


adrifdez (approved):

AlbertoElias (approved):

src/views/transactions/components/transaction-overview/transaction-overview.view.tsx
```diff
@@ -170,16 +170,19 @@ function TransactionOverview({
...
         <Spinner />
-        <p className={classes.signingText}>Sign in your connected wallet to confirm transaction</p>
+        <p className={classes.signingText}>{loaderText}</p>
       </div>
     ) : (
       <PrimaryButton label={buttonLabel} onClick={handleFormSubmit} disabled={isButtonDisabled} />
     );
 
+  const l1TxLoaderText = "Sign in your connected wallet to confirm transaction";
+  const l2TxLoaderText = "Processing transaction";
```
elias-garcia, line 184: The behavior for L2 txs is a bit different. Instead of displaying the spinner with a text, we just need to show nothing but the submit button disabled. The button is going to appear already disabled as you will see in L118 and L121 that we are already disabling it for L2 txs.

PR #745, opened 2021-11-12
prevent propagate a withdrawal done

### LINKED ISSUE #737: WITHDRAW_ALREADY_DONE
After sign one withdraw we can back to sign again and get this error. 

This error is happening as well if you wait after sign the withdraw and if the tx has been confirmed and you are still seen the black box when polling, that error is show

adrifdez:
Closes #737 

### What does this PR does?

Prevent attempt to send a withdrawal confirmed backing to the previous page or checking if it's instant

## Checklist

review them:

### DIFF
--- a/src/store/transactions/withdraw/withdraw.thunks.ts
+++ b/src/store/transactions/withdraw/withdraw.thunks.ts
@@ -219,7 +219,7 @@ function withdraw(
             }
             handleTransactionSuccess(dispatch, account.accountIndex);
           })
-          .catch((error: Error) => {
+          .catch((error) => {
             console.error(error);
             dispatch(withdrawActions.stopTransactionApproval());
             handleTransactionFailure(dispatch, error);
@@ -242,7 +242,7 @@ function withdraw(
             );
             handleTransactionSuccess(dispatch, account.accountIndex);
           })
-          .catch((error: Error) => {
+          .catch((error) => {
             console.error(error);
             dispatch(withdrawActions.stopTransactionApproval());
             handleTransactionFailure(dispatch, error);
--- a/src/store/transactions/withdraw/withdraw.thunks.ts
+++ b/src/store/transactions/withdraw/withdraw.thunks.ts
@@ -1,4 +1,4 @@
-import { replace } from "connected-react-router";
+import { push } from "connected-react-router";
 import { BigNumber } from "ethers";
 import { CoordinatorAPI, Tx, TxFees } from "@hermeznetwork/hermezjs";
 import { getPoolTransactions } from "@hermeznetwork/hermezjs/src/tx-pool";
@@ -180,6 +180,7 @@ function withdraw(
       global: { wallet, signer },
     } = getState();
     const withdrawalId = `${account.accountIndex}${exit.batchNum}`;
+    const alreadyDone = "WITHDRAW_ALREADY_DONE";
 
     dispatch(withdrawActions.startTransactionApproval());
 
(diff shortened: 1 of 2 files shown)


AlbertoElias (dismissed):

src/store/transactions/withdraw/withdraw.thunks.ts
```diff
@@ -252,7 +252,7 @@ function withdraw(
 
 function handleTransactionSuccess(dispatch: AppDispatch, accountIndex: string) {
   dispatch(openSnackbar("Transaction submitted"));
-  dispatch(push(`/accounts/${accountIndex}`));
+  dispatch(replace(`/accounts/${accountIndex}`));
```
amonsosanz, line 255: Shouldn't we consider a more resilient solution? I mean, I can enter the URL manually. Probably not a priority, but it feels like we should keep track of the state of the transaction (or ask the API) and activate the button accordingly.

It feels wrong to interfere with the default behavior of the back button. I think that the back button should send the user back to the previous screen (as expected) and the Withdraw button should be disabled once it has been clicked, as we currently do with the Exit.

src/store/transactions/withdraw/withdraw.thunks.ts
```diff
@@ -252,7 +252,7 @@ function withdraw(
 
 function handleTransactionSuccess(dispatch: AppDispatch, accountIndex: string) {
   dispatch(openSnackbar("Transaction submitted"));
-  dispatch(push(`/accounts/${accountIndex}`));
```
elias-garcia, line 255: This is not going to prevent the user to reach the withdraw from copy & paste a url and clicking on submit, also I don' like the `replace` approach as it breaks the normal flow of the browser.

The ideal flow should be to check if the withdraw has already been submitted by checking if it's still in `localStorage` or something like this, as I think that we clear them once they are mined in L1. But I'm not sure, we can meet on monday to talk about this :)

src/store/transactions/withdraw/withdraw.thunks.ts
```diff
@@ -180,6 +180,7 @@ function withdraw(
       global: { wallet, signer },
     } = getState();
     const withdrawalId = `${account.accountIndex}${exit.batchNum}`;
+    const alreadyDone = "WITHDRAW_ALREADY_DONE";
```
amonsosanz, line 183: Why write a message to be shown to the user uppercase and using underscores?

amonsosanz: Well, I know, it's like an error code. I would prefer a standard msg, but it's ok.

amonsosanz (approved):

src/store/transactions/withdraw/withdraw.thunks.ts
adrifdez: the initial idea was show directly the message but comes with a lot of irrelevant info for the user, so I used the same string to match and show

src/store/transactions/withdraw/withdraw.thunks.ts
```diff
@@ -217,10 +218,15 @@ function withdraw(
             }
             handleTransactionSuccess(dispatch, account.accountIndex);
           })
-          .catch((error) => {
+          .catch((error: Error) => {
             console.error(error);
             dispatch(withdrawActions.stopTransactionApproval());
-            handleTransactionFailure(dispatch, error);
+
+            if (error?.message.indexOf(alreadyDone)) {
```
elias-garcia, line 225: I would move this logic to the `handleTransactionFailure` function, as we already have there the new logic that @amonsosanz did. I suggest this:

```ts
function handleTransactionFailure(dispatch: AppDispatch, error: unknown) {
  const withdrawAlreadyDoneErrorCode = "WITHDRAW_ALREADY_DONE";
  const errorMsg = persistence.getErrorMessage(error);
  const snackbarMsg = errorMsg.includes(withdrawAlreadyDoneErrorCode)
    ? "The withdraw has already been done"
    : errorMsg;

  dispatch(openSnackbar(`Transaction failed - ${snackbarMsg}`, theme.palette.red.main));
}
```

Doing this we take advantage of the `persistence.getErrorMessage` function and we display a better message to the end user instead just `WITHDRAW_ALREADY_DONE`, which it's quite strange for an end user.

elias-garcia (changes requested):

elias-garcia (approved):

AlbertoElias (approved):

PR #744, opened 2021-11-12
Move all error decoding and msg extraction logic to persistence

amonsosanz:
Closes #739.

### What does this PR does?

Move all error decoding and msg extraction logic to persistence to fix a bug found when the signatures are rejected from metamask.

## Checklist

These are the criteria that every PR should meet, please check them off as you
review them:

- [x] Respect code style and lint
- [ ] Update documentation (if needed)

PR #746, opened 2021-11-15
Refactor to Typescript slice 12 (Force Exit)

elias-garcia:
### What does this PR does?

This PR refactors the force exit flow to Typescript.

### How to test?

1. Do a force exit.

## Checklist

review them:

### DIFF
--- a/src/store/transactions/force-exit/force-exit.actions.ts
+++ b/src/store/transactions/force-exit/force-exit.actions.ts
@@ -34,7 +34,7 @@ export interface GoToChooseAccountStep {
 
 export interface GoToBuildTransactionStep {
   type: ForceExitActionTypes.GO_TO_BUILD_TRANSACTION_STEP;
-  account: Account | null;
+  account: Account;
 }
 
 export interface GoToReviewTransactionStep {
@@ -102,7 +102,7 @@ function goToChooseAccountStep(): GoToChooseAccountStep {
   };
 }
 
-function goToBuildTransactionStep(account: Account | null): GoToBuildTransactionStep {
+function goToBuildTransactionStep(account: Account): GoToBuildTransactionStep {
   return {
     type: ForceExitActionTypes.GO_TO_BUILD_TRANSACTION_STEP,
     account,
--- a/src/store/transactions/force-exit/force-exit.actions.ts
+++ b/src/store/transactions/force-exit/force-exit.actions.ts
@@ -1,0 +1,190 @@
+import { BigNumber } from "@ethersproject/bignumber";
+
+// domain
+import { Account, PoolTransaction } from "src/domain/hermez";
+import { Accounts } from "src/persistence";
+
+export enum ForceExitActionTypes {
+  GO_TO_CHOOSE_ACCOUNT_STEP = "[FORCE EXIT] GO TO CHOOSE ACCOUNT STEP",
+  GO_TO_BUILD_TRANSACTION_STEP = "[FORCE EXIT] GO TO BUILD TRANSACTION STEP",
+  GO_TO_REVIEW_TRANSACTION_STEP = "[FORCE EXIT] GO TO REVIEW TRANSACTION STEP",
+  LOAD_ACCOUNTS = "[FORCE EXIT] LOAD ACCOUNTS",
+  LOAD_ACCOUNTS_SUCCESS = "[FORCE EXIT] LOAD ACCOUNTS SUCCESS",
+  LOAD_ACCOUNTS_FAILURE = "[FORCE EXIT] LOAD ACCOUNTS FAILURE",
+  LOAD_POOL_TRANSACTIONS = "[FORCE EXIT] LOAD POOL TRANSACTIONS",
+  LOAD_POOL_TRANSACTIONS_SUCCESS = "[FORCE EXIT] LOAD POOL TRANSACTIONS SUCCESS",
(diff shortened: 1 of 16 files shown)


src/store/transactions/force-exit/force-exit.thunks.ts
```diff
@@ -0,0 +1,108 @@
...
+  };
+}
+
+function handleTransactionSuccess(dispatch: AppDispatch) {
+  dispatch(openSnackbar("Transaction submitted"));
+  dispatch(push("/"));
+}
+
+function handleTransactionFailure(dispatch: AppDispatch, error: Error | string) {
+  const errorMsg = error instanceof Error ? error.message : error;
```
amonsosanz, line 104: When we cancel the TX from Metamask, `error` is not an instance of `Error`. Metamask sends an object with a field `message` which is not an `Error`.

This is a mistake we have been spreading and that I've solved in the PR [#744](https://github.com/hermeznetwork/wallet-ui/pull/744). We should somehow include this case when both branches are merged.

You can easily see the problem in your branch by rejecting a force-exit  from Metamask and looking at the error:

image

src/views/transactions/components/fee/fee.view.jsx
```diff
@@ -26,7 +25,6 @@ function Fee({
   preferredCurrency,
   fiatExchangeRates,
   showInFiat,
-  tokensPriceTask,
```
amonsosanz, line 29: Since `tokensPriceTask` is not required, we should remove it from `src/views/transactions/components/transaction-form/transaction-form.view.jsx` when we instantiate the `Fee` component.

adrifdez (approved):

src/views/transactions/force-exit/force-exit.view.tsx
```diff
@@ -0,0 +1,257 @@
...
+      return {
+        type: "page",
+        data: {
+          title: "Force withdraw",
+          goBackAction: forceExitActions.goToBuildTransactionStep(),
+          closeAction: push(""),
+        },
+      };
+    }
+    default: {
```
amonsosanz, line 224: Same here. We have 3 steps so we should not require these defaults.

src/views/transactions/force-exit/force-exit.view.tsx
```diff
@@ -0,0 +1,257 @@
...
+                    fiatExchangeRatesTask.status === "reloading"
+                      ? fiatExchangeRatesTask.data
+                      : {}
+                  }
+                />
+              )
+            );
+          }
+          default: {
+            return <></>;
```
amonsosanz, line 172: I don't think this is required, `step` should be `never` here.

elias-garcia: Right! I've fixed it in all the flows.

src/store/transactions/force-exit/force-exit.thunks.ts
amonsosanz: One option is for you to approve my PR first so I can merge it and then you can use the helper in this PR.

src/views/transactions/components/fee/fee.view.jsx
elias-garcia: Right! I've removed it from all the flows also

src/store/transactions/force-exit/force-exit.reducer.ts
```diff
@@ -0,0 +1,154 @@
...
+  ForceExitAction,
+  ForceExitActionTypes,
+  Step,
+  TransactionToReview,
+} from "src/store/transactions/force-exit/force-exit.actions";
+import { AsyncTask } from "src/utils/types";
+// domain
+import { Account, PoolTransaction } from "src/domain/hermez";
+import { EstimatedWithdrawFee } from "src/domain";
+import { AccountsWithPagination } from "../transfer/transfer.reducer";
```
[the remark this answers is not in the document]
elias-garcia: Right, leftover from copy and paste 😎

src/store/transactions/force-exit/force-exit.actions.ts
```diff
@@ -0,0 +1,190 @@
...
+  to: Partial<Account>;
+}
+
+export interface GoToChooseAccountStep {
+  type: ForceExitActionTypes.GO_TO_CHOOSE_ACCOUNT_STEP;
+}
+
+export interface GoToBuildTransactionStep {
+  type: ForceExitActionTypes.GO_TO_BUILD_TRANSACTION_STEP;
+  account?: Account;
```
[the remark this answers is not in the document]
elias-garcia: Right, changed to `null` to avoid having the `undefined`. In fact, we need to change some properties from some reducers that could be `undefined` to `null` also. I'll try to do it in another PR.

src/store/transactions/force-exit/force-exit.thunks.ts
elias-garcia: Yes, we can merge your PR later and then I'll bring the changes to this branch.

src/store/transactions/force-exit/force-exit.actions.ts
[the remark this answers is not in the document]
amonsosanz: Wait, why null? The `account` should also not be null in the payload of the action. It should be always an `Account`.

amonsosanz (approved):

PR #750, opened 2021-11-16
solve problem with USD null

### LINKED ISSUE #685: Wallet breaks if token.USD === null when accessing Tx Form for that token
If in the API request to /tokens, one of the tokens has a USD value of null (or undefined), the wallet has an uncaught error in Tx Form. We should just show -- in the fiat values instead of breaking completely.

adrifdez:
Closes #685 

### What does this PR does?

Cover a scenario where the app breaks if you reach tx form and apply some logic to not allow transfers or exits if we don't receive usd value

## Checklist

review them:

### DIFF
--- a/src/views/shared/amount-input/amount-input.view.jsx
+++ b/src/views/shared/amount-input/amount-input.view.jsx
@@ -159,27 +159,29 @@ function AmountInput(Component) {
      * into account (if applicable).
      */
     function handleSendAll() {
-      const maxPossibleAmount = BigNumber.from(account.balance);
-      const maxAmountWithoutFee = getMaxTxAmount(
-        transactionType,
-        maxPossibleAmount,
-        account.token,
-        fee,
-        gasPrice
-      );
+      if (isFinite(fee)) {
+        const maxPossibleAmount = BigNumber.from(account.balance);
+        const maxAmountWithoutFee = getMaxTxAmount(
+          transactionType,
+          maxPossibleAmount,
+          account.token,
+          fee,
+          gasPrice
+        );
 
-      const maxAmountWithoutFeeInFiat = trimZeros(convertAmountToFiat(maxAmountWithoutFee), 2);
+        const maxAmountWithoutFeeInFiat = trimZeros(convertAmountToFiat(maxAmountWithoutFee), 2);
 
-      if (showInFiat) {
-        setValue(maxAmountWithoutFeeInFiat.toString());
-      } else {
-        const newValue = getFixedTokenAmount(maxAmountWithoutFee, account.token.decimals);
+        if (showInFiat) {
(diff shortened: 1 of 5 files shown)


src/views/transactions/components/transaction-form/transaction-form.view.jsx
```diff
@@ -154,6 +154,11 @@ function TransactionForm({
   function isContinueDisabled() {
     if (!isAmountInputDirty) {
       return true;
+    } else if (
+      account.token.USD === null &&
+      (transactionType === TxType.Transfer || transactionType === TxType.Exit)
```
AlbertoElias, line 159: It should show the correct error

adrifdez: Do you mean to show an alert with an explication? The user doesn't reach this page in a normal flow

AlbertoElias: They could be in Account Details and click Transfer. This would skip the Account Selector, right?

adrifdez: You don't see the button in account details neither in the selector if the usd comes as null

AlbertoElias (approved):

src/views/shared/amount-input/amount-input.view.jsx
```diff
@@ -159,27 +159,29 @@ function AmountInput(Component) {
...
     function handleSendAll() {
-      const maxPossibleAmount = BigNumber.from(account.balance);
-      const maxAmountWithoutFee = getMaxTxAmount(
-        transactionType,
-        maxPossibleAmount,
-        account.token,
-        fee,
-        gasPrice
-      );
+      if (isFinite(fee)) {
```
amonsosanz, line 162: Same as [here](#r755008869).

src/views/transactions/components/account-selector/account-selector.view.jsx
```diff
@@ -68,6 +68,9 @@ function AccountSelector({
                     </div>
                   );
                 } else {
+                  const accounts = accountsTask.data.accounts.filter(
+                    (account) => account.token.USD !== null
+                  );
```
amonsosanz, line 73: Since `token.USD` can not be `null` at the domain level, this filter makes no sense. We can do this now only because this view is not yet migrated to TS.

amonsosanz:
The exception (app break) should have been fixed in [this PR](https://github.com/hermeznetwork/wallet-ui/pull/795/files#diff-119553cba46b6e3b670cbbdc90775303e05395a624ded0cbbe1800862be26d18R30).

Please try to reproduce the error before fixing it, since now the property `Token.USD` is already nullable at the domain level.

Make sure though we're showing `--` as this issue describes. Previous PR does not guarantee that.

PR #768, opened 2021-11-26
add recoverPendingDelayWithdraw

### LINKED ISSUE #760: Investigate multidevice with pendingDelayedWithdraw
If we clean localStorage or change to another device while we've an pendingDelayedWithdraw waiting for last step, we can't complete the whole process. We need to check if the api return some value that maybe help us to recovery the lose step. `delayedWithdrawRequest` is returning a value if the exit was signed to be delayed

adrifdez:
Closes #760 

### What does this PR does?

Add a new method to recover pendingDelayedWithdraws from the network if it's not in the localStorage 

### How to test?

Create a pendingDelayedWithdraw and clear pendingDelayedWithdraws in localStorage or change to incognito

## Checklist

review them:

### DIFF
--- a/src/store/account-details/account-details.thunks.ts
+++ b/src/store/account-details/account-details.thunks.ts
@@ -8,6 +8,7 @@ import { AppState, AppDispatch, AppThunk } from "src/store";
 import * as ethereum from "src/utils/ethereum";
 import { createAccount } from "src/utils/accounts";
 import * as accountDetailsActions from "src/store/account-details/account-details.actions";
+import * as globalThunks from "src/store/global/global.thunks";
 // domain
 import {
   HermezAccount,
@@ -272,7 +273,10 @@ function fetchExits(tokenId: Token["id"]): AppThunk {
     } = getState();
     if (wallet !== undefined) {
       return CoordinatorAPI.getExits(wallet.hermezEthereumAddress, true, tokenId)
-        .then((exits: Exits) => dispatch(accountDetailsActions.loadExitsSuccess(exits)))
+        .then((exits: Exits) => {
+          dispatch(globalThunks.recoverPendingDelayedWithdrawals(exits));
+          dispatch(accountDetailsActions.loadExitsSuccess(exits));
+        })
         .catch((err: Error) => dispatch(accountDetailsActions.loadExitsFailure(err)));
     }
   };
--- a/src/store/global/global.thunks.ts
+++ b/src/store/global/global.thunks.ts
@@ -456,9 +456,6 @@ function recoverPendingDelayedWithdrawals(exits: Exits): AppThunk {
       if (chainId !== undefined) {
         const { hermezEthereumAddress } = wallet;
         const provider = Providers.getProvider();
-        const {
-          data: { chainId },
-        } = ethereumNetworkTask;
         const accountPendingDelayedWithdraws: PendingDelayedWithdraw[] =
           storage.getPendingDelayedWithdrawsByHermezAddress(
             pendingDelayedWithdraws,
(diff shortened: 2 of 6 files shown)


src/store/account-details/account-details.thunks.ts
```diff
@@ -272,7 +273,10 @@ function fetchExits(tokenId: Token["id"]): AppThunk {
     } = getState();
     if (wallet !== undefined) {
       return CoordinatorAPI.getExits(wallet.hermezEthereumAddress, true, tokenId)
-        .then((exits: Exits) => dispatch(accountDetailsActions.loadExitsSuccess(exits)))
+        .then((exits: Exits) => {
+          dispatch(globalThunks.recoverPendingDelayedWithdrawals(exits));
```
AlbertoElias, line 277: Shouldn't this be an account details thunk?

src/store/global/global.thunks.ts
```diff
@@ -440,6 +441,69 @@ function checkPendingDelayedWithdrawals(): AppThunk {
   };
 }
 
+/**
+ * Checks if we have some exit that is not saved in localStorage
+ * @returns {void}
```
AlbertoElias, line 446: No param needed with typescript

src/store/global/global.thunks.ts
```diff
@@ -440,6 +441,69 @@ function checkPendingDelayedWithdrawals(): AppThunk {
   };
 }
 
+/**
+ * Checks if we have some exit that is not saved in localStorage
```
AlbertoElias, line 445: It seems like we should just remove the storage for this and depend entirely on this method? And then in the exit component or the getExits thunk we filter between delayed withdraws and normal withdraws. Not sure what the best place would be, maybe @amonsosanz @elias-garcia have thoughts here

src/store/global/global.thunks.ts
```diff
@@ -440,6 +441,69 @@ function checkPendingDelayedWithdrawals(): AppThunk {
...
+          if (
+            !batchNumPendingDelayedWithdraws.includes(exit.batchNum) &&
+            exit.delayedWithdrawRequest
+          ) {
+            try {
+              const blockWithTransactions = await provider.getBlockWithTransactions(
+                exit.delayedWithdrawRequest
+              );
+              const pendingDelayedWithdraw = blockWithTransactions.transactions.find(
+                (t) => Addresses.getEthereumAddress(hermezEthereumAddress) === t.from
```
AlbertoElias, line 483: tx is the normal short form for transaction

src/store/account-details/account-details.thunks.ts
adrifdez: It's global like `addPendingDelayedWithdraw` because it's used in home and account details

src/store/global/global.thunks.ts
adrifdez: At this moment we need something while the tx is confirmed in L1 and keep localStorage give us some consistency until find a better solution. I think that we could improve the flow based in the information from the network

src/store/global/global.thunks.ts
```diff
@@ -440,6 +441,68 @@ function checkPendingDelayedWithdrawals(): AppThunk {
...
+          storage.getPendingDelayedWithdrawsByHermezAddress(
+            pendingDelayedWithdraws,
+            chainId,
+            hermezEthereumAddress
+          );
+        const batchNumPendingDelayedWithdraws = accountPendingDelayedWithdraws.map(
+          (account) => account.batchNum
+        );
+
+        exits.exits.map(async (exit) => {
```
amonsosanz, line 472: It feels strange using `map` here since no transformation is required. I mean, the result of this map is a list of promises that we don't return. I think that using `forEach` feels more correct.

src/store/global/global.thunks.ts
amonsosanz: In general terms, I kind of agree with @AlbertoElias here in the sense that we should not rely on the local-store version of something that we are already obtaining from the API. That said, TBH I don't understand the reasoning from @adrifdez here:

> we need something while the tx is confirmed in L1

src/store/global/global.thunks.ts
```diff
@@ -440,6 +441,68 @@ function checkPendingDelayedWithdrawals(): AppThunk {
...
+          if (
+            !batchNumPendingDelayedWithdraws.includes(exit.batchNum) &&
+            exit.delayedWithdrawRequest
+          ) {
+            try {
```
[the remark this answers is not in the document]
adrifdez: `getBlockWithTransactions` could fail requesting the transactions

src/store/global/global.thunks.ts
adrifdez: I know but I prefer use map instead of for of because we can't use async inside forEach

src/store/global/global.thunks.ts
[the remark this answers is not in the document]
amonsosanz: It feels strange to me using `async` and then having to wrap the code in a `try/catch`. A promise already provides a `try/catch` around it. If an exception happens in `getBlockWithTransactions` it would be caught and treated as a rejection.

It's probably a style issue...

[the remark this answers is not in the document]
amonsosanz: Merging this comment with the other one about the `map` vs `forEach` this looks simpler and more correct to me:

```typescript
exits.exits.forEach((exit) => {
  if (
    !batchNumPendingDelayedWithdraws.includes(exit.batchNum) &&
    exit.delayedWithdrawRequest
  ) {
    void provider
      .getBlockWithTransactions(exit.delayedWithdrawRequest)
      .then((blockWithTransactions) => {
        const pendingDelayedWithdraw = blockWithTransactions.transactions.find(
          (tx) => Addresses.getEthereumAddress(hermezEthereumAddress) === tx.from
        );
        if (pendingDelayedWithdraw) {
          dispatch(
            addPendingDelayedWithdraw({
              ...exit,
              hash: pendingDelayedWithdraw.hash,
              id: `${exit.accountIndex}${exit.batchNum}`,
              hermezEthereumAddress: wallet.hermezEthereumAddress,
              isInstant: false, // TODO I'll remove this key that it's unused with the ExitCard refactor
              timestamp: new Date(blockWithTransactions.timestamp * 1000).toISOString(),
            })
          );
        }
      });
  }
});
```

* No map, since we're not transforming any value
* No try/catch since the promise already provides it
* No need for `async/await`

Does it make sense to you?

amonsosanz (approved):

PR #777, opened 2021-12-07
Refactor to typescript slice 17 (qrcode and myaccount)

### LINKED ISSUE #772: Refactor to Typescript 17
MyAccount and QR

adrifdez:
Closes #772 
### What does this PR does?

Refactor of `my-account` and `qr-code` views

## Checklist

review them:


package.json
```diff
@@ -47,6 +47,7 @@
     "zod": "^3.8.2"
   },
   "devDependencies": {
+    "@types/qrcode.react": "^1.0.2",
```
elias-garcia, line 50: This dependency should be in the `dependencies` object instead of the `devDependencies` one.

src/views/my-code/my-code.styles.ts
```diff
@@ -1,8 +1,10 @@
 import { createUseStyles } from "react-jss";
 
-import { MY_CODE } from "../../constants";
+import { Theme } from "src/styles/theme";
```
elias-garcia, line 3: You can remove the empty below this import.

src/hermezjs-typings.d.ts
```diff
@@ -780,7 +780,7 @@ declare module "@hermeznetwork/hermezjs/src/environment" {
 
   function isEnvironmentSupported(chainId: number): boolean;
 
-  // function getBatchExplorerUrl();
+  function getBatchExplorerUrl(): string;
```
elias-garcia, line 783: Why uncommenting this function? I think that you are not using it anywhere in this PR, right?

src/utils/currencies.ts
```diff
@@ -4,27 +4,32 @@ import { getTokenAmountString } from "@hermeznetwork/hermezjs/src/utils";
...
+  code: string;
+  symbol: string;
+};
+
 const CurrencySymbol = {
   USD: {
     symbol: "$",
     code: "USD",
-  },
+  } as Currency,
```
elias-garcia, line 16: Instead of doing this `as Currency` all the time is better to type the const like this:

```ts
const CurrencySymbol: Record<string, Currency> = // ...
```

src/hermezjs-typings.d.ts
adrifdez: It's used in `my-account.view` line 132

elias-garcia (approved):

PR #779, opened 2021-12-07
Refactor to typescript slice 18 (shared components)

### LINKED ISSUE #775: Refactor to TypeScript slice 18
This slice is focused on refactoring to Typescript the remaining components under `src/views/shared`:

* `account`
* `base-layout`
* `main-header`
* `page-header`
* `portal`
* `private-layout`
* `private-route`
* `public-layout`
* `public-route`
* `route`
* `sidenav`
* `snackbar`
* `spinner`
* `transaction-actions`
* `under-maintenance-error`

It will also include the following dependant refactors:

* `routing/routes.tsx`

amonsosanz:
Closes #775.

### What does this PR does?

This PR is focused on refactoring to Typescript the remaining components under `src/views/shared`:

* `account`
* `base-layout`
* `main-header`
* `page-header`
* `portal`
* `private-layout`
* `private-route`
* `public-layout`
* `public-route`
* `route`
* `sidenav`
* `snackbar`
* `spinner`
* `transaction-actions`
* `under-maintenance-error`

It will also include the following dependant refactors:

* `routing/routes.tsx`

## Checklist

review them:


src/views/shared/private-layout/private-layout.view.tsx
```diff
@@ -1,8 +1,12 @@
...
 
-import usePrivateLayoutStyles from "./private-layout.styles";
+import usePrivateLayoutStyles from "src/views/shared/private-layout/private-layout.styles";
 
-function PrivateLayout({ children }) {
+interface PrivateLayoutProps {
+  children: JSX.Element;
+}
+
+function PrivateLayout({ children }: PrivateLayoutProps): JSX.Element {
```
elias-garcia, line 9: I would suggest to declare it as:

```ts
const PrivateLayout: React.FC = ({ children }) => { //... }
```

Doing this you are going to inherit the `children` from the React interface, which I think is better.

src/views/shared/public-layout/public-layout.view.tsx
```diff
@@ -1,13 +1,18 @@
...
+import { ReactComponent as HermezLogoAlternative } from "src/images/hermez-logo-alternative.svg";
+import { Theme } from "src/styles/theme";
 
-function PublicLayout({ children }) {
-  const theme = useTheme();
+interface PublicLayoutProps {
+  children: JSX.Element;
+}
+
+function PublicLayout({ children }: PublicLayoutProps): JSX.Element {
```
elias-garcia, line 14: I would suggest the same as before to avoid declaring the `children` prop explicitly.

src/views/shared/portal/portal.view.tsx
```diff
@@ -1,19 +1,22 @@
...
-import usePortalStyles from "./portal.styles";
 
-export const PortalSelector = {
-  FULLSCREEN_MODAL: "#fullscreen-modal-root",
-  SIDENAV: "#sidenav-root",
-};
+import usePortalStyles from "src/views/shared/portal/portal.styles";
 
-function Portal({ selector = PortalSelector.FULLSCREEN_MODAL, children }) {
+export type PortalSelector = "#fullscreen-modal-root" | "#sidenav-root";
```
elias-garcia, line 6: The idea here should be to abstract the selector from the parent components. It's kinda ugly to see this from a parent component: `<Portal selector="#fullscreen-modal-root" />`. Maybe what we can do here is to rename the `selector` property to `target = "full-screen" | "sidenav"`, and then based on this target prop pick one selector or another to use internally in this component.

elias-garcia (changes requested):

src/views/shared/private-layout/private-layout.view.tsx
amonsosanz: Ok. Done.

src/views/shared/public-layout/public-layout.view.tsx
amonsosanz: Done.

src/views/shared/portal/portal.view.tsx
amonsosanz: Yes, much better. Done.

src/views/shared/portal/portal.view.tsx
```diff
@@ -1,19 +1,19 @@
 import React from "react";
 import ReactDOM from "react-dom";
-import usePortalStyles from "./portal.styles";
 
-export const PortalSelector = {
-  FULLSCREEN_MODAL: "#fullscreen-modal-root",
-  SIDENAV: "#sidenav-root",
-};
+import usePortalStyles from "src/views/shared/portal/portal.styles";
+interface PortalProps {
```
elias-garcia, line 5: Missing an empty line before the interface.

adrifdez (approved):

PR #781, opened 2021-12-14
Fix Remaining Time to include hours

### LINKED ISSUE #780: Fix Remaining Time to include hours
We currently show the remaining time for a batch to be forged in minutes. We now do this every 6h, so we should show 3h and 30m for example

amonsosanz:
Closes #780

### What does this PR does?

Formats TX pending time as '2h and 23m' or just '23m'.

## Checklist

review them:


src/utils/transactions.ts
```diff
@@ -60,6 +60,15 @@ function getTxPendingTime(
   return timeLeftToForgeInMinutes > 0 ? timeLeftToForgeInMinutes : 0;
 }
 
+/**
+ * Formats a number of minutes to a string as "2h and 23m" or just "23m"
+ */
+function formatMinutes(minutes: number): string {
+  const remMin = Math.round(minutes % 60);
```
adrifdez, line 67: I think that it's better `ceil` than `round` to not show 0 min when minutes are less than 0.5

amonsosanz: Done!

adrifdez (approved):

elias-garcia (approved):

PR #783, opened 2021-12-14
Refactor to typescript slice 19

### LINKED ISSUE #776: Refactor to Typescript slice 19
Refactor home and account details views

adrifdez:
Closes #776 

### What does this PR does?

Refactor to TS of views:
`Home`
`Account details`

## Checklist

review them:


src/store/account-details/account-details.thunks.ts
```diff
@@ -142,8 +142,8 @@ function filterExitsFromHistoryTransactions(
 // ToDo: Define fromItem type
 function fetchHistoryTransactions(
   accountIndex: HermezAccount["accountIndex"],
-  fromItem: number,
-  exits: Exits
+  exits: Exits,
+  fromItem?: number
```
amonsosanz, line 146: Since you have changed the order of these params, now the function `mapDispatchToProps` on `src/views/account-details/account-details.view.tsx` is wrongly naming the params.

src/utils/fees.ts
```diff
@@ -131,7 +131,7 @@ function getEstimatedWithdrawFee(
     fiatExchangeRates
   );
 
-  return exitFeeInFiat + withdrawFeeInFiat;
+  return exitFeeInFiat && withdrawFeeInFiat && exitFeeInFiat + withdrawFeeInFiat;
```
AlbertoElias, line 134: I think this should now be abstracter to variables to make it easier to read

adrifdez: I'm not seeing more variables, maybe for be more verbose we can add the if with the follow and we cover zero as well
```
if (exitFeeInFiat === undefined || withdrawFeeInFiat === undefined) {
    return undefined;
  }
  return exitFeeInFiat + withdrawFeeInFiat;
```

AlbertoElias: As long as undefined is the only falsy value, sure. another option would be `!!exitFeeInFiat === false`

adrifdez: I've pushed it with `undefined` that I think it's more exact

src/views/account-details/account-details.view.tsx
```diff
@@ -54,34 +120,48 @@ function AccountDetails({
...
+    ? storage.getTimerWithdrawsByHermezAddress(
+        timerWithdraws,
+        ethereumNetworkTask.data.chainId,
+        wallet.hermezEthereumAddress
+      )
+    : [];
 
   React.useEffect(() => {
-    onChangeHeader(accountTask.data?.token.name);
+    if (accountTask.status === "successful") {
```
amonsosanz, line 162: isAsyncTaskDataAvailable?

src/utils/currencies.ts
```diff
@@ -41,7 +41,7 @@ function isValidCurrencySymbolKey(key: string): key is CurrencySymbolKey {
 /**
  * Gets the string representation of a token amount with fixed decimals
  */
-function getFixedTokenAmount(amount: string, decimals = 18): string {
+function getFixedTokenAmount(amount: string | undefined, decimals = 18): string {
```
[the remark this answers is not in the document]
adrifdez: Because I didn't remove this TODO https://github.com/hermeznetwork/wallet-ui/blob/0c79527637534cfe5b1532b0fa3864b0a2189c16/src/utils/transactions.ts#L22, I do it now and remove this change

src/views/account-details/account-details.view.tsx
```diff
@@ -227,11 +316,13 @@ function AccountDetails({
             }
```
amonsosanz, line 316: Can't we assume that we also want to show the Spinner when these async tasks are `pending`?

If so, we could do:

```typescript
if (
  !isAsyncTaskDataAvailable(accountTask) ||
  !isAsyncTaskDataAvailable(poolTransactionsTask) ||
  !isAsyncTaskDataAvailable(historyTransactionsTask) ||
  !isAsyncTaskDataAvailable(exitsTask)
) {
  return <Spinner />;
}

if (
  wallet &&
  isAsyncTaskDataAvailable(coordinatorStateTask) &&
  isAsyncTaskDataAvailable(fiatExchangeRatesTask)
) {...}
```

adrifdez: I thought about that but maybe pending is not a case that we want to show the spinner because is not going to change to loading until other thing happens, I'm not sure about it

src/store/global/global.thunks.ts
```diff
@@ -105,11 +105,8 @@ function fetchFiatExchangeRates(): AppThunk {
       .then((fiatExchangeRates: FiatExchangeRates) =>
         dispatch(globalActions.loadFiatExchangeRatesSuccess(fiatExchangeRates))
       )
-      .catch(() => {
-        // ToDo: How are we returning simulated Fiat exchange rates when this request fails???
-        dispatch(
-          globalActions.loadFiatExchangeRatesSuccess(priceUpdaterApi.mockedFiatExchangeRates)
-        );
+      .catch((err: Error) => {
```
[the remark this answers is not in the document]
adrifdez: My fault, I took reference from wrong place https://github.com/hermeznetwork/wallet-ui/blob/8a0b4d0ee231246dae6525d61b04c9194c6a7f95/src/store/global/global.thunks.ts#L914

src/views/home/home.view.tsx
```diff
@@ -0,0 +1,470 @@
...
+  isHermezAccount,
+  PendingDeposit,
+  PoolTransaction,
+  TimerWithdraw,
+} from "src/domain/hermez";
+import { EthereumNetwork } from "src/domain/ethereum";
+
+type HomeStateProps = HomeState &
+  Pick<
+    GlobalState,
```
amonsosanz, line 46: In the case you agreed with [this comment](#r769494184), we should change it also here.

src/views/account-details/account-details.view.tsx
```diff
@@ -1,30 +1,96 @@
...
+  PendingDeposit,
+  PoolTransaction,
+  TimerWithdraw,
+  Token,
+} from "src/domain/hermez";
+import { EthereumNetwork } from "src/domain/ethereum";
+import { AppDispatch, AppState } from "src/store";
+import { Exits } from "src/persistence";
+
+type AccountDetailsStateProps = AccountDetailsState &
```
amonsosanz, line 47: I don't think it's a good idea to intersect here with `AccountDetailsState`. The shape of `AccountDetailsState` could differ from `AccountDetailsStateProps` in the future.

src/views/home/home.view.tsx
```diff
@@ -0,0 +1,470 @@
...
+  FiatExchangeRates,
+  HermezWallet,
+  isHermezAccount,
+  PendingDeposit,
+  PoolTransaction,
+  TimerWithdraw,
+} from "src/domain/hermez";
+import { EthereumNetwork } from "src/domain/ethereum";
+
+type HomeStateProps = HomeState &
```
amonsosanz, line 44: I don't think it's a good idea to intersect here with `HomeState`. The shape of `HomeState` could differ from `HomeStateProps` in the future.

src/views/account-details/account-details.view.tsx
adrifdez: The idea is that they should be the same because we create a `store` for these views and we are bring that data directly from there

src/views/home/home.view.tsx
adrifdez: The idea is that they should be the same because we create a `store` for these views and we are bring that data directly from there

src/views/home/home.view.tsx
```diff
@@ -0,0 +1,470 @@
...
+    onLoadExits,
+    onCheckPendingWithdrawals,
+    onCheckPendingDelayedWithdrawals,
+  ]);
+
+  React.useEffect(() => {
+    if (
+      pendingDepositsCheckTask.status === "successful" &&
+      poolTransactionsTask.status === "successful" &&
+      fiatExchangeRatesTask.status === "successful" &&
```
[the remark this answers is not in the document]
adrifdez: I prefer not to use it inside of useEffect

src/views/account-details/account-details.view.tsx
adrifdez: @elias-garcia let us know your opinion in this case and HomeState

src/utils/currencies.ts
```diff
@@ -53,12 +53,13 @@ function getFixedTokenAmount(amount: string, decimals = 18): string {
 function getAmountInPreferredCurrency(
   usdAmount: number,
   preferredCurrency: string,
-  fiatExchangeRates: FiatExchangeRates
-): number {
+  fiatExchangeRates?: FiatExchangeRates
```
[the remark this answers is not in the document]
amonsosanz: Since there's a case where the `FiatExchangeRates` are not used (when `preferredCurrency === CurrencySymbol.USD.code`) we agreed on leaving `fiatExchangeRates` param as optional.

src/utils/currencies.ts
```diff
@@ -69,8 +70,8 @@ function getTokenAmountInPreferredCurrency(
   amount: string,
   usdTokenExchangeRate: number,
   preferredCurrency: string,
-  fiatExchangeRates: FiatExchangeRates
-): number {
+  fiatExchangeRates?: FiatExchangeRates
```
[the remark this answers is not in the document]
amonsosanz: Same as [here](#discussion_r770306481).

src/utils/currencies.ts
```diff
@@ -91,8 +92,8 @@ function convertTokenAmountToFiat(
   tokenAmount: string,
   token: Token,
   preferredCurrency: string,
-  fiatExchangeRates: FiatExchangeRates
-): number {
+  fiatExchangeRates?: FiatExchangeRates
```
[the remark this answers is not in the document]
amonsosanz: Same as [here](#discussion_r770306481).

src/utils/fees.ts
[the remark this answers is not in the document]
amonsosanz: Same as [here](#discussion_r770306481).

src/views/account-details/account-details.view.tsx
amonsosanz: Yeah, that makes total sense.

src/views/account-details/account-details.view.tsx
amonsosanz: We agreed to proceed with this change for 2 reasons:

* All the async tasks are launched at page load
* The `pending` status was rendering a fragment (now it renders the spinner)

src/views/account-details/account-details.view.tsx
amonsosanz: > The idea is that they should be the same because we create a `store` for these views

Yeah, it's true that we have a slice of the store for each view but still, it somehow feels weird to me to use that slice of the store to intersect with other props of the view. I recognize though that it's a debatable detail, probably biased by my personal taste. Let's see what @elias-garcia thinks.

The question that arises now from my OCD alert system is: If we accept this pattern, should we apply it to all views? Currently, it's only present `AccountDetails` and `Home`.

src/views/home/home.view.tsx
amonsosanz: Same as [here](#discussion_r770323895).

src/views/home/home.view.tsx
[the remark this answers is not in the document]
amonsosanz: Agreed. You're right.

The reasoning is that we do not want asynchronous tasks to be launched when they are in the `reloading` state, to avoid re-fetching data which is already fetching.

src/views/shared/exit-card/exit-card.view.tsx
```diff
@@ -185,7 +188,10 @@ function ExitCard({
         return `${minutes}m`;
       }
     } else {
-      const delayedTime = coordinatorState?.withdrawalDelayer.withdrawalDelay * 1000;
+      if (!coordinatorState) {
+        return "--";
+      }
```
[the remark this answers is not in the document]
amonsosanz: Beautiful!

amonsosanz (dismissed):

src/views/shared/transaction-actions/transaction-actions.view.tsx
```diff
@@ -9,11 +9,11 @@ import { ReactComponent as TokenSwapIcon } from "src/images/icons/token-swap.svg
...
-  hideSend: boolean;
-  hideDeposit: boolean;
-  hideSwap: boolean;
-  accountIndex: string;
-  tokenId: number;
+  tokenId?: number;
+  accountIndex?: string;
+  hideSend?: boolean;
+  hideDeposit?: boolean;
+  hideSwap?: boolean;
```
elias-garcia, line 16: I think that now `hideWithdraw` can also be optional, right?

src/views/shared/fiat-amount/fiat-amount.view.tsx
```diff
@@ -3,13 +3,14 @@ import React from "react";
 import { CurrencySymbol, formatFiatAmount, isValidCurrencySymbolKey } from "src/utils/currencies";
 
 type FiatAmountProps = {
-  amount?: number;
   currency: string;
+  amount?: number;
+  className?: string;
 };
 
 function FiatAmount({
```
elias-garcia, line 11: Just an error from the past in this function. The signature should be:

```ts
function FiatAmount({
  currency,
  amount,
  className,
}: FiatAmountProps): JSX.Element { //... }
```

Can you update it here please?

src/views/shared/transaction-actions/transaction-actions.view.tsx
adrifdez: Yes but I was thinking that we leave it as required until the moment that we need it as optional

src/views/account-details/components/transaction-amount/transaction-amount.view.tsx
```diff
@@ -0,0 +1,60 @@
...
+}
+
+function TransactionAmount({
+  type,
+  fiatAmount,
+  preferredCurrency,
+  fromAccountIndex,
+  accountIndex,
+}: TransactionAmountProps): JSX.Element {
+  const negativeFiatAmount = fiatAmount && -fiatAmount;
```
elias-garcia, line 21: I would add a prop to the `FiatAmount` to display a negative number instead of creating a negative number.

adrifdez: We've decided to remove any negative number

src/views/shared/amount-input/amount-input.view.tsx
```diff
@@ -177,9 +177,12 @@ function AmountInput(
     function handleSendAll() {
       const maxPossibleAmount = BigNumber.from(account.balance);
       const maxAmountWithoutFee = getMaxTxAmount(txType, maxPossibleAmount, fee);
-      const maxAmountWithoutFeeInFiat = trimZeros(convertAmountToFiat(maxAmountWithoutFee), 2);
+      const fiatAmount = convertAmountToFiat(maxAmountWithoutFee);
```
elias-garcia, line 180: I would rename the `fiatAmount` and avoid nested conditionals:

```ts
const maxAmountWithoutFeeInFiat = convertAmountToFiat(maxAmountWithoutFee);
const fixedMaxAmountWithoutFeeInFiat = maxAmountWithoutFeeInFiat
  ? trimZeros(fiatAmount, 2).toString()
  : undefined;
const newValue = showInFiat
  ? fixedMaxAmountWithoutFeeInFiat
  : getFixedTokenAmount(maxAmountWithoutFee.toString(), account.token.decimals);
```

src/views/shared/transaction-actions/transaction-actions.view.tsx
elias-garcia: It's really optional, but it's not being reflected in the interface. So I would add it also so if anyone use this component in the future he knows that it's optional.

elias-garcia (approved):

elias-garcia (approved):

amonsosanz (approved):

PR #794, opened 2021-12-23
Gas withdraw circuit

### LINKED ISSUE #722: Fix withdraw Gas estimation so it estimates for withdrawCircuit
We recently moved to use withdrawCircuit and the gas estimation is different.

adrifdez:
Closes #722 

### What does this PR does?

Use new function that estimate the gasLimit for a WithdrawCircuit

## Checklist

review them:


src/hermezjs-typings.d.ts
```diff
@@ -484,15 +484,15 @@ declare module "@hermeznetwork/hermezjs/src/tx-fees" {
   import { CallOverrides, BigNumber } from "ethers";
 
   // function estimateDepositGasLimit();
+  // function estimateWithdrawGasLimit();
```
amonsosanz, line 487: I'd keep the typing of the function `estimateWithdrawGasLimit` as it was, since `hermezjs-typings.d.ts` declare the types of the library and this is still a function of the library.

elias-garcia: I agree with Alex here.

elias-garcia: Well, at the end we are only typing those functions that we are using, the other functions are commented.

elias-garcia (approved):

src/hermezjs-typings.d.ts
amonsosanz: Because we haven't had the time to type them. The idea is to add all the missing typings and move this file to the lib so we have TS support there for the integrators. I don't see the point of removing a function that was already typed...

AlbertoElias: I don't think we'll migrate hermezjs at this point with the work that's happening for zkEVM

PR #795, opened 2022-01-03
Refine all the domain interfaces against the api schemas

### LINKED ISSUE #774: Refine all the domain interfaces against the API schemas
This task is about fixing all the inconsistencies between the types we've defined in the module `hermezjs-typings.d.ts` and the definition of the schemas provided by the [API docs](https://apidoc.hermez.network/#model-ItemId).

For example, we have this definition:

```typescript
type HistoryTransaction = HermezApiResourceItem & {
  batchNum: number;
  fromAccountIndex: string;
  fromHezEthereumAddress: string;
  id: string;
  toHezEthereumAddress: string | null;
  type: TxType;
  amount: string;
  // fromBJJ: string;
  historicUSD: number | null;
  L1Info: L1Info | null;
  L1orL2: "L1" | "L2";
  L2Info: L2Info | null;
  // position: number;
  timestamp: ISOStringDate;
  // toAccountIndex: string;
  toBJJ: string | null;
  token: Token;
};
```
In our type `fromAccountIndex` is typed as `string` but the [API declares that this value can be returned as `null`](https://apidoc.hermez.network#model-HistoryTransaction):

image

It is necessary to go through all the properties of all the entities defined in the typings file and fix all the inconsistencies and the derived compilation errors.

amonsosanz:
Closes #774

### What does this PR does?

This task is about fixing all the inconsistencies between the types we've defined in the module `hermezjs-typings.d.ts` and the definition of the schemas provided by the [API docs](https://apidoc.hermez.network/#model-ItemId).

## Checklist

review them:


src/views/shared/transaction-info/transaction-info.view.tsx
```diff
@@ -191,9 +201,11 @@ function TransactionInfo({
           from={myHermezAddress}
           to={{
             subtitle: "My Ethereum address",
-            value: getPartiallyHiddenEthereumAddress(
-              getEthereumAddress(transaction.fromHezEthereumAddress)
-            ),
+            value: transaction.fromHezEthereumAddress
```
AlbertoElias, line 204: Maybe abstract this?

amonsosanz: Now that you mention it and I look at the 4 cases, why are we calling `getEthereumAddress` on 3 of the 4 cases?

amonsosanz: I see, it's to remove the `hez:` prefix from the address on L1 TXs.

AlbertoElias (approved):

src/domain/index.ts
```diff
@@ -28,6 +26,12 @@ export type {
   Token,
 } from "@hermeznetwork/hermezjs";
 
+export type HermezNetworkStatus = "online" | "offline";
```
elias-garcia, line 29: This should be `NetworkStatus` as it refers only to the wallet, the `HermezNetworkStatus` is just below this

amonsosanz: You're right! Good cath! Done.

AlbertoElias (approved):

elias-garcia (approved):

PR #798, opened 2022-01-11
Move to persistence all functions that interact with APIs

### LINKED ISSUE #796: Move to persistence all functions that interact with APIs
This task aims to move all the calls to HermezJS functions that interact with the API to the persistence layer of the application. 

The objective is that persistence is the only layer of the application that calls functions that interact with the API, so that later we can easily decode all the data obtained from the API in a centralized place.

This means that any thunk, util, or view that is currently calling a function that does a request to the API uses the persistence layer as a proxy.

> ℹ️ **If the amount of code to be modified is excessive, this issue can be divided into two slices.**

amonsosanz:
Closes #796.

### What does this PR does?

This task aims to move all the calls to HermezJS functions that interact with the API to the persistence layer of the application. 

The objective is that persistence is the only layer of the application that calls functions that interact with the API, so that later we can easily decode all the data obtained from the API in a centralized place.

This means that any thunk, util, or view that is currently calling a function that does a request to the API uses the persistence layer as a proxy.

## Checklist

review them:


src/store/transactions/deposit/deposit.thunks.ts
```diff
@@ -68,33 +66,31 @@ function fetchAccounts(fiatExchangeRates: FiatExchangeRates, preferredCurrency:
...
-              dispatch(depositActions.loadEthereumAccountsSuccess(metaMaskTokens))
-            )
-            .catch((err) => dispatch(depositActions.loadEthereumAccountsFailure(err)));
-        }
-      );
+      return persistence.getTokens(undefined, undefined, undefined, undefined, 2049).then((res) => {
+        ethereum
+          .getTokens(wallet, res.tokens)
+          .then((ethereumAccounts): EthereumAccount[] => {
+            return ethereumAccounts.map((ethereumAccount) => {
```
elias-garcia, line 73: I think that now would make sense to move this this logic to the `ethereum` persistence layer right?

src/types/hermezjs.d.ts
```diff
@@ -624,10 +618,17 @@ declare module "@hermeznetwork/hermezjs/src/api" {
   // function getCreateAccountAuthorization();
   // function getConfig();
 
-  function getPoolTransactions(
-    accountIndex: string,
-    publicKeyCompressedHex: string
-  ): Promise<HistoryTransaction[]>;
+  // function getPoolTransactions(
```
elias-garcia, line 621: Why do we have two `getPoolTransactions` declarations?

elias-garcia (commented):
I propose a change related to the folder structure and naming of the `persistence` folder. Following the hexagonal architecture, `persistence` is just a type of `adapter`. So I propose to rename the `persistence` folder to `adapters`.

I would also remove the folders inside the directory and leave it like this (at the end we need an adapter per service, so I think that it makes more sense):

```
|-- adapters
|      |-- hermez-web.ts
|      |-- price-updater.ts
|      |-- local-storage.ts
|      |-- parsers
|      |      |-- index.ts
```

And the last thing, I would move the code in `/persistence/index.ts` to something like `/persistence/hermez-api.ts`.

src/persistence/index.ts
```diff
@@ -60,7 +90,336 @@ export function postCreateAccountAuthorization(
...
+          TxUtils.getFeeValue(Number(pendingTransaction.fee), pendingTransaction.amount)
+        )
+      );
+    });
+  }
+
+  return totalBalance.toString();
+}
+
+function updateAccountToken(
```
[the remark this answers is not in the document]
amonsosanz: Done.

src/store/account-details/account-details.reducer.ts
```diff
@@ -1,8 +1,7 @@
 import { Pagination } from "src/utils/api";
 // domain
-import { HermezAccount, HistoryTransaction, PoolTransaction } from "src/domain";
+import { Exits, HermezAccount, HistoryTransaction, PoolTransaction } from "src/domain";
 // persistence
```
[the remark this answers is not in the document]
amonsosanz: Removed.

src/store/transactions/deposit/deposit.thunks.ts
amonsosanz: Done. I think it makes total sense. The change was not straight forward though. Please carefully check the commit 0f312c166c77fa0d5ed7bf77abdf0843c1a89306 since I had to:

* Add the params `fiatExchangeRates` and `preferredCurrency` to some functions.
* Make several changes in `getEthereumAccounts` implementation.

Let's make sure I did not break anything... 😓

src/types/hermezjs.d.ts
amonsosanz: Because the library has two implementations, one in the [module `api`](https://github.com/hermeznetwork/hermezjs/blob/main/src/api.js#L152) and another in the [module `tx-pool`](https://github.com/hermeznetwork/hermezjs/blob/main/src/tx-pool.js#L35).

elias-garcia: Oh right! I forgot about the old implementation in the `tx-pool` module hehe

src/store/transactions/deposit/deposit.thunks.ts
amonsosanz: > 

By the way, do you think that `src/utils/ethereum.ts` belongs to persistence? I concluded it does not, but since you refer to it as "the `ethereum` persistence layer" we better discuss it.

amonsosanz: Ethereum util moved to adapters as agreed.

elias-garcia (approved):

AlbertoElias (approved):

PR #804, opened 2022-01-18
Show blocking errors to the user and enable a mechanism to easily report them

### LINKED ISSUE #802: Show blocking errors to the user and enable a mechanism to easily report them

amonsosanz:
Closes #802.

### What does this PR does?

Currently, the Wallet gets stuck showing spinners like this when some errors occur:

These spinners are shown in several places of the app, whenever a response from the API is not available or can not be properly decoded and the data can not be shown.

This task aims to fix this bad UX by showing a generic error message to the user (using a snackbar) that enables her to:

 * **Report**: Open the issue report form with the technical info prefilled.
 * **Close**: Dismiss the error and manually close the snackbar.
 
image

## Checklist

review them:


src/adapters/index.ts
```diff
@@ -19,16 +20,12 @@ const messageKeyErrorParser = StrictSchema<MessageKeyError>()(
...
   if (typeof error === "string") {
     return prefixMsg ? `${prefixMsg}. ${error}` : error;
   } else if (error instanceof ZodError) {
-    const decodingIssues = error.errors
-      .map((issue) => `${issue.message} at path ${issue.path.join(".")}`)
-      .join(" | ");
-    return prefixMsg
-      ? `${prefixMsg}. Decoding issues: ${decodingIssues}`
-      : `Decoding issues: ${decodingIssues}`;
+    const decodingIssues = parseDecodingError(error).join(". ");
```
elias-garcia, line 27: What's the point of this function? I think that we are not going to get any extra information than a regular `JSON.stringify(error)` right?

src/adapters/price-updater.ts
```diff
@@ -46,54 +47,44 @@ const getFiatExchangeRatesResponseParser = StrictSchema<ApiExchangeRateResponse>
...
 /**
  * Returns a list of tokens with usd price.
  */
 function getTokensPrice(): Promise<Token[]> {
-  if (parsedReactAppPriceUpdaterApiUrl.success === false) {
-    return Promise.reject(parsedReactAppPriceUpdaterApiUrl.error.message);
-  } else if (parsedReactAppPriceUpdaterApiKey.success === false) {
-    return Promise.reject(parsedReactAppPriceUpdaterApiKey.error.message);
-  } else {
+  const eitherEnv = adapters.env.getEnv();
```
elias-garcia, line 54: Doing this we are going to parse the `process.env` each time that we need an env var. I would parse it at the beginning of the application and store the result in the `global` store.

src/store/account-details/account-details.thunks.ts
```diff
@@ -243,13 +262,22 @@ function fetchExits(tokenId: Token["id"]): AppThunk {
           dispatch(globalThunks.recoverPendingDelayedWithdrawals(exits));
           dispatch(accountDetailsActions.loadExitsSuccess(exits));
         })
-        .catch((err: unknown) =>
+        .catch((error: unknown) => {
+          const errorMsg = adapters.parseError(
+            error,
+            "An error occurred on src/store/account-details/account-details.thunks.ts:fetchExits"
```
elias-garcia, line 268: I don't think that this is a good idea. What if we rename a file or if we move it to another place?

In fact, this second parameter doesn't provide any relevant information and I would delete it from this`adapters.parseError` function. What if you write this string wrong? A copy and paste for example? Sending `unknown error: JSON.stringify(error)` in this case, which is very unlikely to happen, is more than enough IMO.

src/adapters/env.ts
```diff
@@ -0,0 +1,110 @@
...
+      REACT_APP_ETHERSCAN_URL: z.string(),
+    })
+  )
+);
+
+const developmentEnvParser = StrictSchema<DevelopmentEnv>()(
+  developmentLiteralParser.and(developmentVarsParser)
+);
+
+export function getEnv(): ZodEither<Env, ZodError<Env>> {
```
elias-garcia, line 78: I thought that we agreed on not using monads!

src/domain/index.ts
```diff
@@ -177,3 +177,16 @@ export type ChainTimerWithdraws = Record<HermezEthereumAddress, TimerWithdraw[]>
 
 export type AuthSignatures = Record<ChainId, ChainAuthSignatures>;
 export type ChainAuthSignatures = Record<HermezEthereumAddress, string>;
+
+// User reporting
+export type Message =
```
elias-garcia, line 182: I would call this interface `SnackbarMessage` instead of just `Message`

src/store/global/global.actions.ts
```diff
@@ -443,7 +444,13 @@ function loadFiatExchangeRatesFailure(error: string): LoadFiatExchangeRatesFailu
   };
 }
 
-function openSnackbar(message: string, backgroundColor?: string): OpenSnackbar {
+function openSnackbar({
```
elias-garcia, line 447: All the actions across the app receive params instead of an object, I would keep it like that for consistency.

src/store/global/global.thunks.ts
```diff
@@ -964,7 +1027,73 @@ function fetchTokensPrice(): AppThunk {
...
+              raw: error,
+              parsed: errorMsg,
+            },
+          })
+        );
+      });
+  };
+}
+
+enum ReportErrorFormEntries {
```
elias-garcia, line 1049: I would move this to a `const` inside the `constants.js` file, so we can find and replace the values easily

src/views/shared/snackbar/snackbar.view.tsx
```diff
@@ -1,33 +1,71 @@
...
+  message: Message;
   backgroundColor?: string;
   onClose: () => void;
+  onReport: (raw: unknown, parsed: string) => void;
 }
 
-function Snackbar({ message, backgroundColor, onClose }: SnackbarProps): JSX.Element {
-  const classes = useSnackbarStyles({ backgroundColor });
+function Snackbar({ message, backgroundColor, onClose, onReport }: SnackbarProps): JSX.Element {
+  const bkgColor = backgroundColor
```
elias-garcia, line 21: Importing the `theme` directly from the `src/styles/theme` is not a good idea. I would send the `message.type` as a param in the `useSnackbarStyles` and then decide the final background color inside the styles file: `useSnackbarStyles({ type, backgroundColor })`.

But now that we have 3 clear snackbar states: `info`, `success`, `error` maybe we can remove this `backgroundColor` and just decide it using the `message.type` prop.

src/views/shared/base-layout/base-layout.styles.ts
```diff
@@ -0,0 +1,17 @@
+import { createUseStyles } from "react-jss";
+
+import { Theme } from "src/styles/theme";
+
+const useBaseLayoutStyles = createUseStyles((theme: Theme) => ({
+  spinner: {
```
elias-garcia, line 6: Maybe `spinnerContainer` is better for this class

package.json
```diff
@@ -10,12 +10,14 @@
     "generate-contract-types": "./generate-contract-types.sh",
     "postinstall": "npm run generate-contract-types",
     "start:dev": "react-scripts start",
-    "start:prod": "serve -s -C -p 80 build"
+    "start:prod": "serve -s -C -p 80 build",
+    "touch": "find src/ -type f -name '*.ts*' -exec touch {} +"
```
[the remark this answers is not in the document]
amonsosanz: I thought It would be useful! I will keep using my alias then... Done.

src/adapters/env.ts
```diff
@@ -0,0 +1,110 @@
...
+  const parsedEnvLiteral = z
+    .union([productionLiteralParser, developmentLiteralParser])
+    .safeParse(process.env);
+
+  if (parsedEnvLiteral.success) {
```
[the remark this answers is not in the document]
amonsosanz: The `union` that you propose was also my initial approach but I had to change it because it has a big drawback in terms of error reporting: When the decoding of the union fails you get a list with the errors from all the members of the union.

Let's say you have the 5 vars from development commented (as usual in production). If you have a production variable wrong, then you get 6 errors. This was a no go IMO so I decided to first decode the env var `REACT_APP_ENV` to decide what decoder to use later and produce meaningful error messages.

> The interesting thing here is that you have the `env vars` wrong.

What is wrong? I'm not sure I follow you here.

src/adapters/index.ts
amonsosanz: Human readability.

With this function you get this:
```
Decoding issue: Required at path node. Decoding issue: Expected object, received null at path network. Decoding issue: Expected number, received string at path recommendedFee.createAccount
```

Instead of this:
```
{"issues":[{"code":"invalid_type","expected":"object","received":"undefined","path":["node"],"message":"Required"},{"code":"invalid_type","expected":"object","received":"null","path":["network"],"message":"Expected object, received null"},{"code":"invalid_type","expected":"number","received":"string","path":["recommendedFee","createAccount"],"message":"Expected number, received string"}],"name":"ZodError"}
```

src/adapters/price-updater.ts
amonsosanz: Right, but parsing the `process.env` is so cheap that I don't see the point of storing it already parsed. Store it implies extra code and complexity that we don't need for such a cheap parser.

src/store/account-details/account-details.thunks.ts
amonsosanz: Yes, you're right, this requires maintenance.

> doesn't provide any relevant information

I think it provides relevant information. When you are faced with an error from a customer on production, how will you know on what module and function the error originated? Don't you think this is useful data? Or maybe we can get that info from somewhere else I'm not counting on?

src/domain/index.ts
amonsosanz: I called this `SnackbarMessage` initially when it was tightened to the snackbar, but now that it's exported from the domain, I don't think it's a good idea to couple it with the snackbar. This models a message to show to the user. It can be shown using a snackbar, a modal window, a dialog, or whatever visual component.

src/store/global/global.actions.ts
amonsosanz: Right. Done.

src/store/global/global.thunks.ts
amonsosanz: Yeah makes sense. Done.

src/views/shared/snackbar/snackbar.view.tsx
amonsosanz: > But now that we have 3 clear snackbar states: `info`, `success`, `error`

Well, we currently have 2 states: `info` and `error`.

> maybe we can remove this `backgroundColor` and just decide it using the `message.type` prop.

This has some implications. Currently, we are showing a red/green snackbar on the connection lost/restored events. These are not errors, but info messages. If we proceed with your proposal of inferring the color from the type of the message, these snackbars will become white (or whatever bkg color we decide for `info` messages).

I agree with the change though, should I proceed?

src/views/shared/base-layout/base-layout.styles.ts
amonsosanz: Done!

src/adapters/index.ts
elias-garcia: But what's the point of having to maintain 50 lines of code just to add `Decoding issue` and `received`? For us, the developers (who are going to read that code) is better to have a JSON than that string in one line.

In that example I prefer to go to a JSON Formatter and in 5 seconds get this:

```js
{
   "issues":[
      {
         "code":"invalid_type",
         "expected":"object",
         "received":"undefined",
         "path":[
            "node"
         ],
         "message":"Required"
      },
      {
         "code":"invalid_type",
         "expected":"object",
         "received":"null",
         "path":[
            "network"
         ],
         "message":"Expected object, received null"
      },
      {
         "code":"invalid_type",
         "expected":"number",
         "received":"string",
         "path":[
            "recommendedFee",
            "createAccount"
         ],
         "message":"Expected number, received string"
      }
   ],
   "name":"ZodError"
}
```

src/domain/index.ts
elias-garcia: Okay, good argument!

src/store/account-details/account-details.thunks.ts
elias-garcia: It provides relevant information only:

- If you write the path and the name of the function right.
- If you don't modify the name of the function or of the file or even if you don't move the file in the future.

We can not get this info from somewhere as this info is only in the `Error` class (`error.stack`). But if this is not supported by the language we cannot go too far, maintaining this strings is going to be a hell compared to the useful info that it's going to provide.

src/adapters/price-updater.ts
elias-garcia: Well, the extra code is just 3 redux actions and 1 thunk... I prefer to do this just one when the app loads instead of having to parse it each time that i need to access the env

src/views/shared/snackbar/snackbar.view.tsx
elias-garcia: Yep, but the idea here would be to introduce the `success` state to be able to have the green snackbar as well.

src/adapters/env.ts
elias-garcia: I would check if it's exported from `zod` then. I prefer to do that than creating our own monad type for this (and having it explicitly in our code)

src/adapters/index.ts
amonsosanz: Let's wait for a third opinion on this one. I see the value of a human-readable msg but you clearly don't.

src/adapters/env.ts
elias-garcia: Well... In fact, we could only use `parse` instead of `safeParse`, at the end this function should only return the `Env` domain entity.

amonsosanz: I already checked and they don't. It's declared inline:

image

amonsosanz: Wait! they just added the type in the [last version](https://github.com/colinhacks/zod/pull/732/files)! Do you want me to upgrade to [3.11.6](https://github.com/colinhacks/zod/releases/tag/v3.11.6)?

src/views/shared/snackbar/snackbar.view.tsx
amonsosanz: That is not the problem. The problem is that currently the `error` state is designed to process real errors and a connection lost event is not an error. How do you comply with this interface?

```
{
      type: "error";
      text?: string;
      raw: unknown;
      parsed: string;
}

File: src/adapters/env.ts (hunk 4 above)
Comment (elias-garcia) [reply src/adapters/env.ts:78]: I think that this a better approach:

> Well... In fact, we could only use `parse` instead of `safeParse`, at the end this function should only return the Env domain entity.

The adapters should return the domain entities, the same as we are doing in `hermez-api` adapters etc

File: src/views/shared/snackbar/snackbar.view.tsx (hunk 8 above)
Comment (elias-garcia) [reply src/views/shared/snackbar/snackbar.view.tsx:21]: That's why I would like to have an `ErrorDetails` domain entity with all that fields:

```ts
{
  text?: string;
  raw: unknown;
  parsed: string;
}
```

Because it's an Error in the application (network, decoding...) but it has nothing to do with a UI error. The snackbar interface should be something like this:

```
{
  type: "error"; // red
  error: "string" | ErrorDetails;
}
```

The connection lost is an `Error` if we talk about UI, and the snackbar should be able to display it. We need to decouple domain from views always!

(The names of the interfaces, props etc are just an idea)

src/adapters/env.ts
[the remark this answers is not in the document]
amonsosanz: We agreed to keep the current implementation as it provides more accurate errors to the user.

src/adapters/index.ts
amonsosanz: @elias-garcia does not see the point of human-readable errors because as technical errors, they are always going to be interpreted by technicians. We are going to remove the function in charge of parsing the errors to a human-readable version and reporting only the stringified version of the error.

src/store/account-details/account-details.thunks.ts
amonsosanz: Since we're removing the function `parseError` we still need to decide what do we want to store in the state for each errored AsyncTask.

src/views/shared/snackbar/snackbar.view.tsx
amonsosanz: We agreed to go this way to separate "colored messages" from "actual errors" that can be reported:

```
// User reporting
export type Message =
  | {
      type: "info-msg" | "success-msg" | "error-msg"; // white | green | red
      text: string;
    }
  | {
      type: "error";
      text?: string;
      raw: unknown;
      parsed: string;
    };
```

amonsosanz: Done!

src/adapters/env.ts
elias-garcia: We agreed in renaming the `ZodEither` to simply `Either`.

src/store/account-details/account-details.thunks.ts
amonsosanz: We agreed on keeping the function `parseError`:

  * Without `prefixMsg` 
  * Without parsing the decoding issues to a human-readable string
  * Including the serialized error with the stack trace

src/adapters/price-updater.ts
amonsosanz: Ok. I will implement the 3 redux actions and the thunk on global and create a `useEffect` in `app.view`.

src/store/account-details/account-details.thunks.ts
amonsosanz: Done

src/adapters/index.ts
amonsosanz: Done

src/adapters/env.ts
amonsosanz: Done

src/adapters/env.ts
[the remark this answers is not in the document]
AlbertoElias: I tend to prefer simpler approaches, and Elias' one is easier to read. I think it's also enough to just notify that the environment variables are wrong.

Also +1 that production vars are actually shared vars, not specific to production.

src/adapters/price-updater.ts
amonsosanz: Done

src/views/app.view.tsx
```diff
@@ -75,9 +77,10 @@ function App({
   useAppStyles();
 
   React.useEffect(() => {
+    onLoadEnv();
     onCheckHermezStatus();
     onLoadFiatExchangeRates();
```
elias-garcia, line 82: Is this working now? As it's only called once, it could be possible that the `env` is still `undefined`. I think that it's more secure to create another hook and be sure that we have the `env` properly loaded:

```ts
React.useEffect(() => {
  if (env) {
    onCheckHermezStatus(env);
    onLoadFiatExchangeRates(env);
  }
}
```

src/store/transactions/deposit/deposit.thunks.ts
```diff
@@ -157,7 +164,7 @@ function deposit(amount: BigNumber, ethereumAccount: EthereumAccount): AppThunk
     if (wallet !== undefined && signer !== undefined) {
       return adapters.hermezApi
         .deposit(
-          HermezCompressedAmount.compressAmount(amount.toString()),
+          amount,
```
AlbertoElias, line 167: Why change this?

AlbertoElias (approved):

src/views/app.view.tsx
amonsosanz: Yes, it is working, but you have a point there. Changed.

src/store/transactions/deposit/deposit.thunks.ts
amonsosanz: This was a bug. We were already compressing the amount in the [deposit function of the hermez-api adapter](https://github.com/hermeznetwork/wallet-ui/blob/develop/src/adapters/hermez-api.ts#L602).

elias-garcia (approved):

PR #806, opened 2022-01-20
Fix error showing account fiat balances upon refresh

amonsosanz:
### What does this PR does?

This PR fixes a bug that shows the fiat balances of the accounts as `--` after some time since the page has been loaded.

## Checklist

review them:

### DIFF
--- a/src/adapters/hermez-api.ts
+++ b/src/adapters/hermez-api.ts
@@ -25,7 +25,6 @@ import {
   HermezAccount,
   HermezAccounts,
   HermezRawAccount,
-  HermezRawAccounts,
   HermezWallet,
   HistoryTransaction,
   HistoryTransactions,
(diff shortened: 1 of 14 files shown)


AlbertoElias (approved):

src/adapters/hermez-api.ts
```diff
@@ -482,6 +440,48 @@ function createHermezAccount(
   };
 }
 
+function getHermezRawAccounts(
```
elias-garcia, line 443: Instead of having this `getHermezRawAccounts` function I would keep it simple and just keep the the `getHermezAccounts`. I don't like to introduce too much terminology like the `getXXRaw`, we already have the `unknownXXXParser` and we are going to get lost with all this stuff soon...

amonsosanz: Makes total sense. Done!

elias-garcia (approved):

PR #810, opened 2022-01-26
Include some unmanaged errors and avoid reporting denied signature errors

### LINKED ISSUE #809: Include some unmanaged errors and avoid reporting denied signature errors
This issue is mainly focused on two tasks:

* Prevent showing the report error button on user denied signature errors.
* Properly manage some errors that were not considered in the previous slice.

amonsosanz:
Closes #809.

### What does this PR does?

This issue is mainly focused on two tasks:

* Prevent showing the report error button on user denied signature errors.
* Properly manage some errors that were not considered in the previous slice.

## Checklist

review them:


src/store/global/global.thunks.ts
```diff
@@ -895,16 +957,34 @@ function checkPendingDeposits(): AppThunk {
...
+                    type: "error",
+                    raw: error,
+                    parsed: errorMsg,
+                  })
+                );
+              });
           })
-          .catch(() => ({}));
+          .catch((error: unknown) => {
+            const errorMsg = adapters.parseError(error);
```
AlbertoElias, line 972: It really feels like this catch handler can be abstracted. There's a lot of repeated code

AlbertoElias (dismissed):

src/store/global/global.thunks.ts
amonsosanz: Agreed. We should work on this abstraction once we're finished and have the whole picture clear.

amonsosanz: I've installed the lib [`stacktrace-js`](https://www.stacktracejs.com/) to produce meaningful stack traces for the errors using the source maps and as you suggested, abstracted the repetitive code to a global thunk named `processError`.

AlbertoElias (approved):

src/adapters/index.ts
```diff
@@ -20,27 +21,62 @@ const messageKeyErrorParser = StrictSchema<MessageKeyError>()(
...
+}
+
+const metamaskUserRejectedRequestError = StrictSchema<MetamaskUserRejectedRequestError>()(
+  z.object({
+    code: z.literal(4001),
+    message: z.string(),
+  })
+);
+
+export function isMetamaskUserRejectedRequestError(
```
elias-garcia, line 36: I think that we can move this to the `types.ts` file (at the end type guards are related to types) and move the code from this file to a new `adapters/errors.ts` file, having all this stuff in the `index.ts` doesn't make too much sense (I think that we've already talked about this but just wanted to remind it).

src/store/login/login.thunks.ts
```diff
@@ -236,11 +246,20 @@ function postCreateAccountAuthorization(wallet: HermezWallet.HermezWallet): AppT
...
-        console.error(error);
-        const text = adapters.parseError(error);
-        dispatch(loginActions.addAccountAuthFailure(text));
-        dispatch(globalActions.openSnackbar({ type: "info-msg", text }));
-        dispatch(loginActions.goToWalletSelectorStep());
+        adapters
+          .parseError(error)
+          .then((text) => {
+            dispatch(loginActions.addAccountAuthFailure(text));
+            dispatch(
```
elias-garcia, line 253: Why we don't call the `processError` here and in L159 of the same file and we parse the error instead and open the snackbar?

src/store/transactions/withdraw/withdraw.thunks.ts
```diff
@@ -202,22 +190,27 @@ function handleTransactionSuccess(dispatch: AppDispatch, accountIndex: string) {
...
-          }
-    )
-  );
+  if (adapters.isMetamaskUserRejectedRequestError(error) === false) {
+    const withdrawAlreadyDoneErrorCode = "WITHDRAW_ALREADY_DONE";
+    adapters
+      .parseError(error)
+      .then((errorMsg) => {
+        dispatch(
+          openSnackbar(
```
elias-garcia, line 199: Same as before, why we don't call the `processError` here?

src/adapters/index.ts
amonsosanz: > I think that we can move this to the `types.ts` file

Do you mean moving only the type guard?

I thought it made sense to keep the type guard `isMetamaskUserRejectedRequestError`, next to the type `MetamaskUserRejectedRequestError` and the parser `metamaskUserRejectedRequestError`. We could move the type guard to `types.ts` but being such small pieces, don't you see also a benefit on having them all together?

> move the code from this file to a new `adapters/errors.ts `

Done!

> I think that we've already talked about this

We did when we only had one error-related function here, but now it's true that it deserves its own module.

src/store/login/login.thunks.ts
amonsosanz: Because `processError` would always open the error-report snackbar and here we might need to show an `info-msg` to the user. In other words, we only want to show the error-report snackbar when we have an exception or decoding error (`Error` or `ZodError`) but here we may need to show other "non-reportable" issues to the user (as we were doing before).

src/store/transactions/withdraw/withdraw.thunks.ts
amonsosanz: Same answer applies, the message `"The withdraw has already been done"` is not reportable. It's only an "info-message" we show to the user.

src/store/login/login.thunks.ts
elias-garcia: Okay perfect, i like the approach!

src/adapters/index.ts
elias-garcia: > Do you mean moving only the type guard?

I thought it made sense to keep the type guard isMetamaskUserRejectedRequestError, next to the type MetamaskUserRejectedRequestError and the parser metamaskUserRejectedRequestError. We could move the type guard to types.ts but being such small pieces, don't you see also a benefit in having them all together?

I think that is more semantic to have all the stuff related to type checking like guards and types in one place, and then import them from the correct module. For example, I think that it's better to call `types.isMetamaskUserRejectedRequestError` instead of `adapters.errors.isMetamaskUserRejectedRequestError` from the `global.thunks` which is not what you'd expect without taking a look at the code.

amonsosanz: Well, I see the module `src/utils/types.ts` more for helper types as `AsyncTask` or `Either`, but ok, it's not very relevant.

Do you agree on moving the 3 related elements (type, parser and, type guard) or do you prefer moving only the type guard?

elias-garcia: Well, the `isMetamaskUserRejectedRequestError` at the end is a helper that you use across the app (for example in the `global.thunks` and in the `error.adapters`)

amonsosanz: Ok. Done!

elias-garcia (approved):

PR #813, opened 2022-02-02
update logo

mmilenkovic:
### What does this PR does?

Updates logo.

## Checklist

review them:


public/index.html
```diff
@@ -9,7 +9,7 @@
       name="description"
       content="A functional web wallet to interact with the Hermez zk-rollup user accounts."
     />
-    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
+    <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo.png" />
```
AlbertoElias, line 12: is the new logo only 192px wide? If not we should maintain a 192 version and continue to use that one

mmilenkovic: The best we can get from https://drive.google.com/drive/folders/1f-5DJwgRYQSgEr5wtgTj62Hq_xogx5aB is 128px so I'll add that one.

AlbertoElias (approved):

